PR #418, opened 2018-01-16
Remove documentation that we do not support - custom  Auth0

adamtheturtle:
## Description

https://jira.mesosphere.com/browse/DCOS_OSS-2044 - Custom Auth0 documentation does not work (community question)

This removes the documentation that does not work.

I built the documentation locally and on 1.10 and 1.9 things looked as I hoped.
However, I don't know how to view 1.11 documentation locally.

Also, will the build tell us if there are now broken links within the documentation? Is there a good way to check for these?

## Urgency
- [ ] Blocker 
- [X] High
- [ ] Medium

## Requirements


jgehrcke:
Approved. Thanks, Adam. It's sad that we have to do so, but removing this is better than having wrong docs. As of today, we do not have the capacity to debug this and correct this.

CC @somikbehera @artemharutyunyan

PR #476, opened 2018-01-26
Update service statuses

leemunroe:
## Description
In 1.11 we’ve changed how statuses behave in the “Services” tab of the UI and we need to explain these statuses in our docs.

* Updated statuses for 1.11
* Removed link to Marathon docs (no longer applies)

## Urgency
- [ ] Blocker 
- [x] High
- [ ] Medium

## Requirements


pages/1.11/gui/services/index.md
```diff
@@ -23,3 +23,22 @@ Click the service name to open the Instances panel, which provides CPU, memory,
 For services with a web interface, hover over the service name and click ![open service](/1.11/img/open-service.png) to view it.
 
 **Tip:** You can access the Mesos web interface at `<hostname>/mesos`.
+
+# Service Status
+
+DC/OS UI introduces the following concepts to illustrate the possible status of a Service at any point in time:
+
+## Running
+The Service is reported as being successfully running.
```
orlandohohmeier: I'd remove "successfully" as the service is "running" if there is at least one "running" tasks and none of the other statuses match.

orlandohohmeier:
@Poltergeist @weblancaster could you please take a look as well?

pages/1.11/gui/services/index.md
leemunroe: Ok. Took this straight from the old Marathon docs. This hasn't changed right, you're just suggesting we improve it? Double checking. https://mesosphere.github.io/marathon/docs/marathon-ui.html#application-status-reference

weblancaster: I believe that's what @orlandohohmeier is suggesting @leemunroe 

I think we have two options here:
1 - Keep what Lee took from Marathon documentation which is correct but vague if anyone wants to understand more.
2 - Keep the same plus add some clarification on how it works.

Running status only happens when there are no instances deploying nor recovering nor stopped in a service (this is not a suggested copy but only an explanation how it works)

pages/1.11/gui/services/index.md
```diff
@@ -23,3 +23,22 @@ Click the service name to open the Instances panel, which provides CPU, memory,
...
+
+## Deploying
+Whenever the user requests a change to the Service, DC/OS is performing the required actions, which haven’t completed yet.
+
+If a Service is having difficulty getting to a Running state, a warning icon is displayed. This indicates the Service is waiting for the resource offers it requires to run or that too many tasks have failed in a short amount of time. More information can be obtained from the “Debug” section of the Service.
+
+## Recovering
+Whenever a change to the Service has been requested by DC/OS (for example if a health check has failed consecutively), the task is killed and a new task is started. Similar to Deploying, DC/OS is performing the required actions, which haven’t completed yet.
```
weblancaster: I would remove the `(for example if a health check has failed consecutively)` because this is a basically a waiting stage.

**Should we use tasks or instances?** I believe we should go with instances because we use the UI and also in code.

pages/1.11/gui/services/index.md
leemunroe: `The Service is reported as running and no instances are reported as deploying or recovering.` Better?

pages/1.11/gui/services/index.md
leemunroe: Good point. In this scenario we should stick to instances. Will remove the health check example.

pages/1.11/gui/services/index.md
```diff
@@ -23,3 +23,22 @@ Click the service name to open the Instances panel, which provides CPU, memory,
 For services with a web interface, hover over the service name and click ![open service](/1.11/img/open-service.png) to view it.
 
 **Tip:** You can access the Mesos web interface at `<hostname>/mesos`.
+
+# Service Status
+
+DC/OS UI introduces the following concepts to illustrate the possible status of a Service at any point in time:
+
+## Running
+The Service is reported as running and no instances are reported as deploying or recovering.
```
weblancaster: I think looks good, what do you think about `The Service is reported as running and no instances are deploying or recovering` though?

orlandohohmeier:
L👀ks good!

PR #539, opened 2018-02-02
added phantomjs pdf build

fatz:
## Description
added phantomjs to build pdfs. Local build takes about 25min now

## Urgency
- [ ] Blocker 
- [ ] High
- [x] Medium


aayvazyan:
Tested, it works. Total Build time 33Min, from which ~20 min was pdf build.

PR #540, opened 2018-02-02
Improve security docs [1.10,1.11]

gpaul:
## Description
This PR adds some help text to the Permissions Reference pages. It also mentions that a necessary permission must be granted when configuring tasks to run as a custom linux user.

If anyone feels it necessary to modify the 1.8 or 1.9 pages I'm happy to do so, too.

https://jira.mesosphere.com/projects/DCOS/issues/DCOS-18669

## Urgency
- [ ] Blocker 
- [ ] High
- [x] Medium

## Requirements

### DIFF
--- a/pages/1.10/security/ent/perms-reference/index.md
+++ b/pages/1.10/security/ent/perms-reference/index.md
@@ -33,6 +33,12 @@ By convention, `full` indicates that the permission supports all other action id
 
 ## <a name="admin-router"></a>Admin Router Permissions
 
+In each of these cases the principal is the client performing the HTTP
+request. The principal is authenticated using the DC/OS Authentication Token it
+presents in the `Authorization` HTTP header. Once the principal has been
+authenticated Admin Router checks that it has been assigned the necessary
+permission to enact the action (eg., `full`) on the resource (eg., `dcos:adminrouter:acs`).
+
 |                                                                                                                                 Permission string                                                                                                                                 | full | C | R | U | D |
 |-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------|---|---|---|---|
 | `dcos:adminrouter:acs`<br>Controls access to the security and access management features.                                                                                                                                                                                         | x    |   |   |   |   |
@@ -55,17 +61,36 @@ By convention, `full` indicates that the permission supports all other action id
 
 ## <a name="mesos"></a>Mesos Permissions
 
+It is important to note that the principal issuing the command to Mesos is the
+one that must be authorized to do so. This is not always the DC/OS user that is
+logged into the UI or CLI. For example, when a DC/OS user uses the CLI or UI to
+launch a Marathon application the Marathon framework performs its own
+authorization of that user before deciding to create the application. Once
+Marathon has determined that the DC/OS user is permitted to create the
+application, Marathon uses its own DC/OS service account user called
+`dcos_marathon` to instruct Mesos to launch the actual Mesos tasks. At that
+point, Mesos will check that the `dcos_marathon` service account (and not the
+end user) is authorized to perform a `create` action on the
+`dcos:mesos:master:task:app_id` resource.
+
+Permission string sections for `<role-name>`s in square brackets are optional.
+These sections allow the permission scope to be narrowed.
+When no role name is specified, the permissions apply to all [Mesos roles](/1.10/overview/concepts/#mesos-role).
+For example, the permission string `dcos:mesos:agent:framework:role` controls view access to DC/OS services registered with any [Mesos role](/1.10/overview/concepts/#mesos-role), whereas the permission string `dcos:mesos:agent:framework:role:slave_public` controls view access to DC/OS services registered with the role `slave_public`.
+
+Applications launched with Root Marathon cannot receive offers with resources reserved for a role which is not `slave_public` or `*`.
+
 |                                                                                                                                 Permission string                                                                                                                                 | full | C | R | U | D |
 |-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------|---|---|---|---|
 | `dcos:mesos:agent:container:app_id[:<service-or-job-group>]`<br> Controls access to the [debugging](/1.10/monitoring/debugging/debug-perms/) features for a specific service or job.                                                                                               |      |   |   | x |   |
-| `dcos:mesos:agent:container:role[:<role-name>]`<br>Controls access to the [debugging](/1.10/monitoring/debugging/debug-perms/) features for a specific role.                                                                                                                       |      |   |   | x |   |
+| `dcos:mesos:agent:container:role[:<role-name>]`<br>Controls access to the [debugging](/1.10/monitoring/debugging/debug-perms/) features for the given [Mesos role](/1.10/overview/concepts/#mesos-role).                                                                                                                       |      |   |   | x |   |
 | `dcos:mesos:agent:endpoint:path[:<endpoint>]`<br> Controls access to unprotected [Mesos endpoints](https://mesos.apache.org/documentation/latest/authorization/).                                                                                                                         |      |   | x |   |   |
 | `dcos:mesos:agent:executor:app_id[:<service-or-job-group>]`<br> Controls view access to service and job [executor information](https://mesos.apache.org/documentation/latest/app-framework-development-guide/).                                                                           |      |   | x |   |   |
 | `dcos:mesos:agent:flags`<br> Controls view access to [agent flag](https://mesos.apache.org/documentation/latest/slave/flags/) configurations.                                                                                                                                              |      |   | x |   |   |
-| `dcos:mesos:agent:framework:role[:<role-name>]`<br> Controls view access to DC/OS services registered with a particular role.                                                                                                                                                     |      |   | x |   |   |
+| `dcos:mesos:agent:framework:role[:<role-name>]`<br> Controls view access to DC/OS services registered with the given [Mesos role](/1.10/overview/concepts/#mesos-role).                                                                                                                                                     |      |   | x |   |   |
 | `dcos:mesos:agent:log`<br>Controls access to the [agent logs](/1.10/monitoring/logging/).                                                                                                                                                                                          |      |   | x |   |   |
 | `dcos:mesos:agent:nested_container_session:app_id[:<service-or-job-group>]`<br> Controls access, by service or job group, to launching a container within a container of a service or job while [debugging](/1.10/monitoring/debugging/).                                          |      | x |   |   |   |
-| `dcos:mesos:agent:nested_container_session:role[:<role-name>]`<br> Controls access, by role, to launching a container within a container of a service or job while [debugging](/1.10/monitoring/debugging/).                                                                       |      | x |   |   |   |
+| `dcos:mesos:agent:nested_container_session:role[:<role-name>]`<br> Controls access, by [Mesos role](/1.10/overview/concepts/#mesos-role), to launching a container within a container of a service or job while [debugging](/1.10/monitoring/debugging/).                                                                       |      | x |   |   |   |
 | `dcos:mesos:agent:nested_container_session:user[:<linux-user-name>]`<br> Controls access, by Linux user, to launching a container within a container of a service or job while [debugging](/1.10/monitoring/debugging/). The users of both nested containers must be the same.     |      | x |   |   |   |
 | `dcos:mesos:agent:sandbox:app_id[:<service-or-job-group>]`<br> Controls access to the Mesos sandbox.                                                                                                                                                                              |      |   | x |   |   |
 | `dcos:mesos:agent:task:app_id[:<service-or-job-group>]`<br> Controls access to task information.                                                                                                                                                                                  |      |   | x |   |   |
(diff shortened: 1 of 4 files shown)


pages/1.10/security/ent/perms-reference/index.md
```diff
@@ -55,37 +61,63 @@ By convention, `full` indicates that the permission supports all other action id
...
+launch a Marathon application the Marathon framework performs its own
+authorization of that user before deciding to create the application. Once
+Marathon has determined that the DC/OS user is permitted to create the
+application, Marathon uses its own DC/OS service account user called
+`dcos_marathon` to instruct Mesos to launch the actual Mesos tasks. At that
+point, Mesos will check that the `dcos_marathon` service account (and not the
+end user) is authorized to perform a `create` action on the
+`dcos:mesos:master:task:app_id` resource.
+
+Permission string sections for `<role-name>`s in square brackets are optional.
```
gpaul: This section about the roles related to `slave_public` was copied from the 1.11 section. It looked true to me, @jgehrcke @vespian can you confirm?

vespian: lgtm

@sambatyon ?

gpaul:
@jongiddy thank you for the thorough review, those were all excellent suggestions and I've integrated or addressed all of them. PTAL

pages/1.10/security/ent/perms-reference/index.md
```diff
@@ -97,13 +128,23 @@ By convention, `full` indicates that the permission supports all other action id
...
 |-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------|---|---|---|---|
 | `dcos:secrets:default:[<path-name>/]<secret-name>`<br> Controls access to individual [secrets](/1.10/security/ent/secrets/).                                                                                                                                                           | x    | x | x | x | x |
 | `dcos:secrets:list:default:/[<path>]`<br> Controls view access to the names of [secrets](/1.10/security/ent/secrets/).                                                                                                                                                                 |      |   | x |   |   |
 
 ## <a name="superuser"></a>Superuser Permissions
 
+Similar to the Windows `Administrator` or Linux `root` accounts, DC/OS has a
+`superuser` role. A user with permission to perform the `full` action on the
```
jgehrcke: Let's not use the term 'role' here. Maybe 'concept'?

Also:

> The special permission dcos:superuser (action identifier irrelevant) overrides everything else: a DC/OS authorizer must never deny an action if this permission is set.

(https://wiki.mesosphere.com/pages/viewpage.action?pageId=113019009)

(`full` is not required as of that spec)

gpaul:
@ahairape would you mind taking a look, too?

pages/1.10/security/ent/perms-reference/index.md
jongiddy: A combination of the original and the updated:
```
Similar to the Windows `Administrator` or Linux `root` accounts, DC/OS has the concept of the `superuser`.
```

pages/1.10/security/ent/perms-reference/index.md
```diff
@@ -33,6 +33,12 @@ By convention, `full` indicates that the permission supports all other action id
 
 ## <a name="admin-router"></a>Admin Router Permissions
 
+In each of these cases the principal is the client performing the HTTP
+request. The principal is authenticated using the DC/OS Authentication Token it
+presents in the `Authorization` HTTP header. Once authenticated, the Admin Router
```
adamtheturtle: We seem to mostly say "Admin Router" rather than "the Admin Router"

cc @jgehrcke @vespian

pages/1.10/security/ent/perms-reference/index.md
gpaul: @jgehrcke I've changed the wording to match what @jongiddy suggested.
I've also changed the `dcos:superuser` permission's columns to have 'x's in all columns and updated the description of what actions are allowed, as of the wiki page you referenced.

pages/1.10/security/ent/perms-reference/index.md
jgehrcke: Agree.

gpaul: Fixed :+1:

pages/1.10/security/ent/perms-reference/index.md
```diff
@@ -55,37 +61,62 @@ By convention, `full` indicates that the permission supports all other action id
...
 
+The required permissions must be assigned to the principal that issues the
+command to Mesos. This is not always the DC/OS user that is logged into the UI
+or CLI. For example, when a DC/OS user uses the CLI or UI to launch a Marathon
+application the Marathon framework performs its own authorization of that user
+before deciding to create the application. Once Marathon has determined that
+the DC/OS user is permitted to create the application, Marathon uses its own
+DC/OS service account user called `dcos_marathon` to instruct Mesos to launch
+the actual Mesos tasks. At that point, Mesos will check that the
+`dcos_marathon` service account (and not the end user) is authorized to perform
```
adamtheturtle: ❓ Maybe clearer if you use "DC/OS user" rather than "end user".

adamtheturtle:
❓ I think that some of this may be better suited to the permissions management page (https://github.com/mesosphere/dcos-docs-site/blob/staging/pages/1.11/security/ent/perms-management/index.md) than the permissions reference.

pages/1.10/security/ent/perms-reference/index.md
```diff
@@ -33,6 +33,12 @@ By convention, `full` indicates that the permission supports all other action id
 
 ## <a name="admin-router"></a>Admin Router Permissions
 
+In each of these cases the principal is the client performing the HTTP
+request. The principal is authenticated using the DC/OS Authentication Token it
```
jgehrcke: I like this paragraph, but I wonder what we gain by using the term 'principal' here .

Am I right that the major goal is to describe how Admin Router associates the HTTP request with a DC/OS user (i.e. DC/OS uid), and then to provide an example of how it evaluates permission data? If that is the case, I have one proposal:

"Most HTTP requests received by Admin Router require authentication proof. After extracting the DC/OS `uid` associated with an incoming HTTP request (by validating and evaluating the DC/OS authentication token presented in the `Authorization` HTTP request header) Admin Router checks that this user has been granted the necessary privilege to perform the requested operation. For example, the DC/OS user identified by said `uid` must have `full` access to the protected resource `dcos:adminrouter:package` in order to be able to access the DC/OS package API through Admin Router."

(I see we use "principal" here: https://docs.mesosphere.com/1.10/security/ent/perms-management/ -- but I think that is more or less an accident, and there we should also introduce "DC/OS user <uid>" or "DC/OS group <gid>".)

gpaul:
> I think that some of this may be better suited to the permissions management page 

I think you're right, but in several places in our docs we link directly to this page and I've personally been confused landing here with no help text or contextual links available. Try `grep -R "perms-reference" pages` for example. A different / future approach would be to link to a Permissions page with links to the other permissions pages.

pages/1.10/security/ent/perms-reference/index.md
gpaul: This clearly shows my struggle to name things consistently in our docs.

pages/1.10/security/ent/perms-reference/index.md
gpaul: That's a beautiful paragraph. I like `principal` because it is a pretty clear term whereas DC/OS user is not. Adam's comment shows how confused all this sounds: 

> "DC/OS user" rather than "end user"

I think there is a distinction between an IAM user account, and someone sitting in from of a computer accessing a DC/OS cluster. In my mind, `DC/OS user` refers most closely to an entry in the `users` table of the IAM. It is short for `DC/OS IAM regular user`. That includes service account users. I tried with 'end user' to imply a physical meaty creature sitting at a computer. Oh my word this is tough ><. 

I'm going to try and get rid of `principal` and just be clear that I mean `DC/OS user`.

gpaul: > Am I right that the major goal is to describe how Admin Router associates the HTTP request with a DC/OS user (i.e. DC/OS uid), and then to provide an example of how it evaluates permission data?

@jgehrcke that, too. My main goal of this PR was to make it clear that when Alice attemps an action which fails, it may very well be because a necessary permission has not been assigned to the `dcos_marathon` or other user and that assigning it to the `alice` won't help. In light of that I wanted it to be clear who the entity being authorized is in each case.

gpaul:
Please see the last commit. I took @jgehrcke's suggestion a little further. If you guys think I should back up a bit that's fine, too. I don't feel strongly about any particular change in this PR but feel strongly about improving the UX on this page. I'm happy to keep iterating on this, I'd rather get it really good than merge and come back to it in a week.

pages/1.10/security/ent/perms-reference/index.md
jgehrcke: "

"Mesos will check that the `dcos_marathon` service user account (and not the DC/OS user account that the HTTP request issued to Marathon is authenticated with) is authorized ..."?

pages/1.10/security/ent/perms-reference/index.md
jgehrcke: I was referring to the goal of this paragraph. :-) 

> I like principal because it is a pretty clear term whereas DC/OS user is not. 

Maybe in this documention not yet. In our concepts it is. Let's change this.

In our concept/architecture there are DC/OS user accounts. Each user account is identified by a corresponding cluster-unique `uid`. There are 'regular user accounts' and 'service user accounts'.

pages/1.11/security/ent/perms-reference/index.md
```diff
@@ -32,13 +33,35 @@ The available actions are `create`, `read`, `update`, `delete`, and `full`.
...
+
+When a HTTP request to a protected resource is received by an authorizer, the
+authorizer inspects the `Authorization` HTTP request header to obtain the DC/OS
+authentication token. The DC/OS authentication token is validated and evaluated
+by the authorizer. Once the `uid` is extracted from the DC/OS authentication
+token the authorizer checks that the corresponding DC/OS user has been granted
+the necessary privilege to perform the requested operation. For example, the
+DC/OS user identified by `uid` must have `full` access to the protected
+resource `dcos:adminrouter:package` in order to be able to access the DC/OS
+package API through Admin Router.
```
jgehrcke: :heart:

pages/1.10/security/ent/perms-reference/index.md
```diff
@@ -55,37 +61,62 @@ By convention, `full` indicates that the permission supports all other action id
...
+or CLI. For example, when a DC/OS user uses the CLI or UI to launch a Marathon
+application the Marathon framework performs its own authorization of that user
+before deciding to create the application. Once Marathon has determined that
+the DC/OS user is permitted to create the application, Marathon uses its own
+DC/OS service account user called `dcos_marathon` to instruct Mesos to launch
+the actual Mesos tasks. At that point, Mesos will check that the
+`dcos_marathon` service account (and not the end user) is authorized to perform
+a `create` action on the `dcos:mesos:master:task:app_id` resource.
+
+Permission string sections for `<role-name>`s in square brackets are optional.
```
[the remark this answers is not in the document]
gpaul: Definitely always true of all square bracket sections.

pages/1.10/security/ent/perms-reference/index.md
```diff
@@ -33,6 +33,12 @@ By convention, `full` indicates that the permission supports all other action id
 
 ## <a name="admin-router"></a>Admin Router Permissions
 
+In each of these cases the principal is the client performing the HTTP
+request. The principal is authenticated using the DC/OS Authentication Token it
+presents in the `Authorization` HTTP header. Once authenticated, Admin Router
+checks that the principal has been assigned the necessary permission to perform
+the action (e.g., `full`) on the resource (e.g., `dcos:adminrouter:acs`).
+
 |                                                                                                                                 Permission string                                                                                                                                 | full | C | R | U | D |
```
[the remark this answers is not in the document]
gpaul: No - we should standardize on resource identifier.

gpaul:
@ahairape thanks for the review, I've addressed your comment. PTAL

gpaul:
@jgehrcke @adamtheturtle @ahairape can we merge this?

pages/1.10/security/ent/perms-reference/index.md
adamtheturtle: @gpaul please can you reply to @jgehrcke 's suggestion. I'd still be OK at this point to ship as-is as it is clearly an improvement and this doc is ripe for conflicts.

pages/1.10/security/ent/perms-reference/index.md
```diff
@@ -97,13 +127,22 @@ By convention, `full` indicates that the permission supports all other action id
...
 | `dcos:secrets:list:default:/[<path>]`<br> Controls view access to the names of [secrets](/1.10/security/ent/secrets/).                                                                                                                                                                 |      |   | x |   |   |
 
 ## <a name="superuser"></a>Superuser Permissions
 
-|                                                                                                                                 Permission string                                                                                                                                 | full | C | R | U | D |
+Similar to the Windows `Administrator` or Linux `root` accounts, DC/OS has the
+concept of the `superuser`. A user with permission to perform any action on the `dcos:superuser` resource has complete, unrestricted access to any operation
```
adamtheturtle: In Slack, our friend @gpaul said:

> in that sense, it [`dcos:superuser`] is not a resource in itself, but a meta resource

Please confirm that "the `dcos:superuser` resource" is what you mean by this.

adamtheturtle:
I will not revoke my approval but consider blocking on @jgehrcke 's preference of not using "principal".

gpaul:
@adamtheturtle it looks like I only changed the wording around `principal` in the 1.11 docs and forgot to perform the same change in the 1.10 docs. I'll do that and merge.

pages/1.10/security/ent/perms-reference/index.md
gpaul: I'm going to stick with 'resource' for now so we don't introduce a new concept (meta-resource) in this document. In future we may have 'security policy' as a concept that more naturally describes  `dcos:superuser`.

gpaul:
@adamtheturtle thanks again for spotting that. I've aligned the 1.10 docs with the 1.11 docs now :+1: 

@jgehrcke there's no more mention of "principal" in this PR. Some resource IDs contain the word, but that is inherited from upstream Apache Mesos which refers to principals.

jgehrcke:
Lovely!

gpaul:
@pavisandhu can we merge this?

jgehrcke:
> @pavisandhu can we merge this?

I have answered this for us :)

Thank you, Gustav.

PR #543, opened 2018-02-02
Fix edge-lb service account permissions

drewkerrigan:
Grant access to the service account for all pools - this has caused confusion for several customers.

### DIFF
--- a/pages/services/edge-lb/0.1.9/installing/index.md
+++ b/pages/services/edge-lb/0.1.9/installing/index.md
@@ -85,18 +85,13 @@ Use the following CLI commands to provision the Edge-LB service account with the
 
 All CLI commands can also be executed via the [IAM API](/1.10/security/iam-api/).
 
-1.  Grant the permissions and the allowed actions to the service account using the following commands. The commands below allow your Edge-LB service to manage DC/OS packages, Marathon tasks, and Edge-LB tasks.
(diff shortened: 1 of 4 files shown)


pages/services/edge-lb/1.0.0/installing/index.md
```diff
@@ -85,18 +85,13 @@ Use the following CLI commands to provision the Edge-LB service account with the
 
 All CLI commands can also be executed via the [IAM API](/1.10/security/iam-api/).
```
masonse: how will this link work once dcos 1.11 comes out? We need to figure out a way where you're not hard coding links to dcos versions.

pages/services/edge-lb/0.1.9/installing/index.md
```diff
@@ -85,18 +85,13 @@ Use the following CLI commands to provision the Edge-LB service account with the
 
 All CLI commands can also be executed via the [IAM API](/1.10/security/iam-api/).
```
masonse: Hardcoded link to a dcos version.  How can we make this link non-dcos version specific (so that it still applies once 1.11 comes out).

pages/services/edge-lb/1.0.0/installing/index.md
drewkerrigan: I believe there is a way to add redirects from /XYZ to /version/XYZ

PR #591, opened 2018-02-20
Update merge scripts to support current dcos-commons structure

nickbp:
Updates the merge scripts for Cassandra, Elastic, HDFS, and Kafka (and their Beta equivalents), which all currently reside in `dcos-commons`. The existing scripts do not work with the current structure in `dcos-commons`. With this update, they are updated to invoke a common `dcos-commons-merge.sh`, which handles the common retrieval and conversion.

To maintain parity with the Jekyll features that are used by the service docs, the following three modules are added to the metalsmith configuration. All three are configured to only be enabled in the `services/` folder, and only take effect in content which explicitly invokes them. See comments in `index.js`.

- `metalsmith-in-place-dcos`: This enables processing of `{{ mustache }}` statements within content. Similar to [Jekyll's built-in support for `{{ mustache.style }}` templating](https://jekyllrb.com/docs/templates/). This `-dcos` version is a rewrite of `metalsmith-in-place`, which uses a `render` annotation in the header, whereas the original was guessing templating types by filename extension. The extension-based check in the original version led to conflicts with other modules which expect a `.md` extension for content, and was otherwise much less flexible.
- `metalsmith-include-content-dcos`: This enables including raw content from other files using C-like `#include path/to/file.txt` statements. Similar to [Jekyll's `{% include %}` statement support](https://jekyllrb.com/docs/includes/). This `-dcos` version is a rewrite of `metalsmith-include-content` which supports explicit filters for where include statements should be processed. The original `metalsmith-include-content` would instead just blindly scan all site content (including images!).
- `metalsmith-data-loader`: This allows fetching data models from external .json and .yml files, to be used when rendering the page. Similar to [Jekyll's `_data` directory support](https://jekyllrb.com/docs/datafiles/). Unlike the above two modules, this one works great as-is.

## Description
Tracked in `INFINITY-3220`.

## Urgency
- [ ] Blocker 
- [ ] High
- [X] Medium

## Requirements


scripts/service-update-scripts/dcos-commons-merge.sh
```diff
@@ -0,0 +1,200 @@
...
+output_package_version_dir=${output_package_dir}/${version}
+if [ -d $output_package_version_dir ]; then
+   echo "Output directory already exists, remove first: $(pwd)/$output_package_version_dir"
+   exit 1
+fi
+
+# Before copying content, add 10 to the sort order of existing index files
+for i in $( ls ${output_package_dir}/*/index.md ); do
+    awk '/^menuWeight:/ {sub(/[[:digit:]]+$/,$NF+10)}1 {print}' $i > $i.tmp && mv $i.tmp $i
+done
```
aayvazyan: ^-- SC2045: Iterating over ls output is fragile. Use globs
https://github.com/koalaman/shellcheck/wiki/SC2045

scripts/service-update-scripts/dcos-commons-merge.sh
```diff
@@ -0,0 +1,200 @@
...
+# Shared content
+
+echo "Updating template content"
+input_template_dir=$REPO_NAME/docs/pages/_includes/services
+output_template_dir=pages/services/include
+
+rm -rf $output_template_dir
+mkdir -p $output_template_dir
+for filepath in $(ls $input_template_dir/*.md); do
+    # e.g. pages/_includes/services/template.md => pages/services/include/template.tmpl
```
aayvazyan: ^-- SC2045: Iterating over ls output is fragile. Use globs
https://github.com/koalaman/shellcheck/wiki/SC2045

scripts/service-update-scripts/dcos-commons-merge.sh
```diff
@@ -0,0 +1,200 @@
...
+rm -f $merge_list_file
+
+###
+# Version-specific content
+
+echo "Copying service content"
+rm -rf ${output_package_version_dir}
+mkdir -p ${output_package_version_dir}
+for filepath in $(ls $input_framework_docs_dir/*.md); do
+    if [ $(basename $filepath) == "index.md" ]; then
```
aayvazyan: rm -rf $REPO_NAME/
       ^-- SC2115: Use "${var:?}" to ensure this never expands to / .

scripts/service-update-scripts/dcos-commons-merge.sh
```diff
@@ -0,0 +1,200 @@
...
+# Update packageName in data.yml (including or omitting beta- prefix)
+sed -i "s/packageName:.*/packageName: $package/g" $output_yml_data_filepath
+
+# Add title/version information to latest index file
+sed -i "s/navigationTitle:.*/navigationTitle: $title $version/g" ${output_package_version_dir}/index.md
+sed -i "s/title:.*/title: $title $version/g" ${output_package_version_dir}/index.md
+
+echo "------------------------------"
+echo " $title $version merge complete"
+echo "------------------------------"
```
aayvazyan: @nickbp can you please run `shellcheck` against this script and make corrections on some "red" comments. I've commented some of them under the exact lines. There are very few major things to correct, the most of suggested corrections are minor things, but this will ensure us that the build job will not make weird issues, infuture.

nickbp: Done. Ended up just cleaning up all the errors.

PR #627, opened 2018-02-28
Pdf

VanessaPC:
## Urgency
- [ ] Blocker 
- [x] High
- [ ] Medium

## Requirements


scripts/mdtopdf.sh
```diff
@@ -0,0 +1,103 @@
...
+        NEW_FILE="${d}/${FILE_NAME}"
+        if [ -f $NEW_FILE ]
+        then
+        # Create temporary file with all md content to send to pandoc // this avoids very long urls & long strings (Pandoc has a limit)
+          TEMP_FILE=$(mktemp)
+          TEMP_FILES="${TEMP_FILES} ${TEMP_FILE}"
+
+          cat ${NEW_FILE} >> ${TEMP_FILE}
+        fi
+      done
```
aayvazyan: can we please change the way how we find files from for loop to somewhat  above ?
 while IFS= read -r -d '' SOURCE_FILE
  do
     ...
     ...
  done <  <(find "${INPUT_FOLDER}" -type f -name "*.md" -print0)

scripts/mdtopdf.sh
```diff
@@ -0,0 +1,103 @@
...
+  echo "======="
+  echo "Starting pdf build $(date)"
+  cat "${PARALLEL_TEMPFILE}" | parallel --halt-on-error 2 --progress --eta --workdir "${PWD}" --jobs "${PARALLEL_JOBS:-4}"
+  echo "Finished build $(date)"
+  echo $TEMP_FILES
+  # rm -f "${TEMP_FILES}"
+}
+
+clean "${OUTPUT_FOLDER}"
+main "${INPUT_FOLDER}" "${OUTPUT_FOLDER}"
```
aayvazyan: Makes sense to run a shellcheck and fix the major issues?
https://github.com/koalaman/shellcheck

scripts/mdtopdf.sh
```diff
@@ -0,0 +1,122 @@
...
+
+     #Fix for all local links to not mislead Pandoc into grabbing other files when there is nothing to look for
+     # issue reported in Pandoc : https://github.com/jgm/pandoc/issues/3619
+     sed -i -r 's,]:,]\\:,g' "${TEMP_FILE}"
+     sed -i -r 's,\/#,,g' "${TEMP_FILE}"
+
+    # Set name for last folder
+    if [ -z "${PDF_FILE_NAME}" ]
+    then
+      "${PDF_FILE_NAME}"="MesosphereDCOS"
```
fatz: i think this should be 

`PDF_FILE_NAME="MesosphereDCOS"` correct?

fatz:
Added a few code comments I was struggling with locally.

Ok some thoughts.

First of all I really like the result. Its a pretty good PDF having TOC and links. Also page numbers and Chapter titles. Its way better than what we had before.

Downside is that the PDF build takes ages (at least on my machine) However that should not stup us. Performance improvements could be done later.

This is IMO not a show stopper as we didn't had this before BUT
- The chapter should be Adjusted. Currently the header line are always of the same value h1, h2 etc.
  - I think as a first improvement we should maybe try to use the yaml "title" as a latex chapter. 
  - A second improvement could be folder ordering by numbers. Like in `https://github.com/progit/progit/tree/master/en`
- Another improvment would be a kind of title page for the version documentation.

So lets say in general LGTM! We should migrate this to dev and take a look into the result. So @pavisandhu could decide if it is ready to be shipped.

scripts/mdtopdf.sh
VanessaPC: Correct, I'll do that change today.

Makefile
```diff
@@ -34,17 +34,11 @@ build-development: build-api
 #
 
 build-pdf-production:
-	./scripts/pdf.sh ./build ./build-pdf
-
-build-pdf-concat-production:
-	./scripts/pdf-concat.sh ./pages ./build-pdf
+	./scripts/mdtopdf.sh ./pages ./build-pdf
 
 build-pdf-development: build-api docker-development-up-pdf
```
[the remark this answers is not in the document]
VanessaPC: These changes have been made in the last commit.

scripts/mdtopdf.sh
VanessaPC: This is also fixed as per the second to last commit.

PR #687, opened 2018-03-08
1.10,1.11/installing/ent/upgrading: wait for cockroachdb underreplicated ranges

gpaul:
## Description
This PR updates the Upgrading documentation for 1.10 and 1.11 to specify that Dan must wait for the underreplicated_ranges value to drop to zero after a master was upgraded. 

@vishnu2kmohan do we have "master replacement" documentation somewhere or are these upgrade docs as close as we have?

Fixes https://jira.mesosphere.com/browse/DOCS-2458

## Urgency
- [ ] Blocker 
- [x] High
- [ ] Medium

## Requirements

### DIFF
--- a/pages/1.10/installing/ent/upgrading/index.md
+++ b/pages/1.10/installing/ent/upgrading/index.md
@@ -187,7 +187,7 @@ Proceed with upgrading every master node one-at-a-time in any order using the fo
     1.  Verify that `curl http://<dcos_master_private_ip>:5050/metrics/snapshot` has the metric `registrar/log/recovered` with a value of `1`.
         **Tip:** If you are upgrading from permissive to strict mode, this URL will be `curl https://...` and you will need a JWT for access.
     1.  Verify that `/opt/mesosphere/bin/mesos-master --version` indicates that the upgraded master is running Mesos 1.4.0.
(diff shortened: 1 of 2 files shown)


surdy:
@gpaul sadly we do not have Master replacement documentation (to the best of my knowledge). Enough customers do master replacement as a regular operation on DC/OS clusters that I think it warrants a docs page on it IMO.

gpaul:
Here are screenshots of the tables for 1.10 and 1.11 using chrome on my laptop:
upgrade_1-10
upgrade_1-11

gpaul:
> @gpaul sadly we do not have Master replacement documentation (to the best of my knowledge). Enough customers do master replacement as a regular operation on DC/OS clusters that I think it warrants a docs page on it IMO.

Given how sensitive the procedures are I'd very much like for us to have clear docs detailing the steps and have automated tests for those exact procedures. If a customer wants to do something even slightly different, we should write respond by telling them to hang on while we add a new test case for their scenario. 

I'm happy to take on this works and may pull info from you and others in #support-team if that's OK.

jgehrcke:
Just noted the

>  **Tip:** If you are upgrading from permissive to strict mode, this URL will be `https://...`.

which was already there previously. How horribly wrong :(.

We should not advertise upgrading from permissive to strict, but more importantly: Exhibitor does not serve HTTPS

PR #735, opened 2018-03-15
Add current Kibana limitation when X-Pack is enabled.

mpereira:


mpereira:
> Is this change applicable to other versions? (and should the same text be added there?)

Yes, it is. I was in the process of doing that, I'm pretty sure it affects all versions, but was still confirming.

> I was under the impression that we had a latest folder here somewhere?

I think `latest` is handled dynamically through templating. Looks like it's happening [here](https://github.com/mesosphere/dcos-docs-site/blob/kibana-xpack-login-limitation/layouts/dcos-docs-landing.pug#L3).

> We discussed over lunch that we would remove the service link -- would this affect this limitation?

Yes it would.

benclarkwood:
The latest folder has not been added yet. I'm working on that this sprint.

PR #743, opened 2018-03-18
Add a note about the need to install Kubernetes

alenkacz:
For more context see my question in #kubernetes channel https://mesosphere.slack.com/files/U7B4ZHXT2/F9SNLEY6T/what_am_i_doing_wrong____.txt

It turned out when you install Kubernetes through UI, it needs to be installed through CLI as well. I think it makes sense to put it explicitly there through this page.


pires:
These changes should be added to mesosphere/dcos-kubernetes repo first. Also, this PR is related to a previous version. The most recent is 1.0.1-1.9.4.

PR #855, opened 2018-04-12
Update template guide and add release notes

drewkerrigan:
This should not be merged until Marathon-LB 1.12 is released.


drewkerrigan:
@asridharan No I'll be adding to this one after 1.12 is tagged

PR #864, opened 2018-04-13
Add note re. event subscriber disabling

edgarlanting:
## Description
Added note regarding disabling event subscriber(s). We've had some customers asking about this and it was not clear how to check this and if needed how to disable this.

## Urgency
- [ ] Blocker 
- [ ] High
- [x] Medium

## Requirements
- Test all commands and procedures. Procedure has been tested.
- Change all affected versions (Add notes to 1.9 docs and remove from 1.10, 1.11 docs).


adyatlov:
Hi @edgarlanting! Thank you for the PR! I think it's a way to go with the DC/OS docs and we have to create more PRs than COPS tickets :). Regarding the change, I have two notes: 

1. It should be `--event_subscriber`, not `--event subscriber`.
2. Marathon team removed this feature in Marathon 1.5 which corresponds to DC/OS 1.10. So for the document you are changing it makes sense to remove the whole list item.

edgarlanting:
fixed the type re. `--event_subscriber`.

PR #918, opened 2018-04-20
COPS-2519 Removing broken links to shell script.

SarahStegall:
## Description
COPS-2519

## Urgency
- [x] Blocker 
- [ ] High
- [ ] Medium

## Requirements


SarahStegall:
I'm not sure why you removed the shell script link from the 1.10 version of this page, and the creators of the 1.11 page are no longer at Mesosphere. In both cases, users will not find the shell script noted, and this is a blocking bug. At Pavi's direction, I have removed both Tips. We can restore them if and when we get a link to a shell script, or if this issue is otherwise fixed.

PR #957, opened 2018-05-01
Changing Overview excerpt

SarahStegall:
## Urgency
- [ ] Blocker 
- [ ] High
- [ X] Medium

## Requirements


SarahStegall:
I have edited all the Overview documents to add excerpts for the title cards. I have also done a light edit pass to standardize headings.

SarahStegall:
Anitha: This is unclear to me.

pages/1.11/overview/design/installation/index.md
Line #9
Trying to treat each of those components as independent artifacts to install and configure on target hosts would introduce failures that would get in the way of relying on the system.

rewrite as:
Trying to treat each of those components as independent artifacts to install and configure on target hosts will introduce failures **that will rely on the system**.

SarahStegall:
Anitha: I liked this a little better:

The bits are identical on every host, with different roles; they just use a different subset of components.

SarahStegall:
I've made the changes suggested by Anitha except where noted. 

I have also discovered that when there is a colon in the title or navigationTitle field in the header, the title card and menu navigation breaks.

SarahStegall:
There's no particular reason to avoid beginning a sentence with "Because", as it is a subordinate clause in this case. Perhaps you are concerned that it would indicate a sentence fragment?

I rewrote the sentence anyway, because that page has too many sentences that start with a subordinate clause, but I don't think we can really make a rule that "we never start a sentence with 'because'." It's a common subordinate conjunction, the same as "before" or "after" or a dozen others.

PR #966, opened 2018-05-03
Ss/docs 2309

SarahStegall:
## Description
https://jira.mesosphere.com/browse/DOCS-2309

## Urgency
- [ ] Blocker 
- [ ] High
- [X] Medium

## Requirements


SarahStegall:
The problem is that the original table was formatting in Markdown using ` markers to denote code. Text formatted as code, whether using ` or <code>, will wrap. I removed the code formatting from all text except that referenced in the Description column, which is buried in text that wraps anyway. The JSON fields referenced in the JSON column now render without wrapping, and maintain sufficient width for the column.

As an experiment, I tried hand-coding this table in HTML to see if I could retain the <code> formatting; the table columns behaved the same as with ` formatting. They wrapped inappropriately and rendered the JSON fields unreadable. 

Corrections to this table were made for 1.9, 1.10, and 1.11. I found no versions of this table in earlier releases, but if you find one let me know and I'll fix it.

PR #1042, opened 2018-05-24
DCOS_OSS-3220 Re-configure 1.9 release notes

SarahStegall:
## Description
https://jira.mesosphere.com/browse/DCOS_OSS-3220

## Urgency
- [ ] Blocker 
- [ ] High
- [x] Medium

Re-ordering/re-configuring release notes to conform to new structure


SarahStegall:
Here are Judith's notes and my response (in boldface). I also added a couple of items at the end:

- [x] In the 1.9.5 section `DCOS-18350 - DC/OS IAM (Bouncer): set TMPDIR to /var/lib/dcos/bouncer/tmp. This allows /tmp, and other directories, to be mounted as noexec.` should have an enterprise tag. (Bouncer is an enterprise only component) 
- [x] For some reason the sections listing changes in 1.9.4 through 1.9.1 are missing. Either we should list all changes since the major release, or **only the ones in this latest release and the features in the major release**. 
- [x]  In the Breaking Changes section, add an enterprise tag to `The DC/OS Identity and Access Management (IAM) SAML service provider implementation no longer accepts transient subject NameIDs.` I believe this is Bouncer, which is an enterprise only component.
- [x]	In the DC/OS internals section is `Update DC/OS internal Python from 3.4 to 3.5.` really enterprise only? I don't have any specific knowledge about the ticket, but it seems weird that it would just be EE. **DCOS-4208 is an Enterprise ticket.**
- [x] In the DC/OS internals section is 'Added dcos-shell which activates the DC/OS environment for running other DC/OS command line tools.` really enterprise only? I don't have any specific knowledge about the ticket, but it seems weird that it would just be EE. **DCOS-27272 is an Enterprise ticket.**
- [x] Is Expanded OS Support really enterprise only? **The links in that line are all to /ent/ pages.**
- [x]	Upgrades Are definitely not enterprise only. Remove this tag.
- [x] Added Enterprise tags to all Jenkins sections in “About 1.9” **(some had them, some didn't)**
- [x]	Corrected “maturity-badge” code in 1.9.0-rc1-rc4 for “experimental” and “preview”

PR #1059, opened 2018-05-31
DCOS-25876 Edits to GCE installation docs for 1.10

SarahStegall:
## Description
https://jira.mesosphere.com/browse/DCOS-25874

The GCE installation procedure for 1.11 is incorrect; the correct version should be based on the GCE installation docs for 1.10. Before we edit 1.11, however, I have gone through and edited the 1.10 docs for clarity. Once the 1.10 docs are correct, we can use them as the basis for the 1.11 corrections.

## Urgency
- [x] Medium

### DIFF
--- a/pages/1.10/installing/oss/cloud/gce/index.md
+++ b/pages/1.10/installing/oss/cloud/gce/index.md
@@ -1,26 +1,25 @@
 ---
 layout: layout.pug
-excerpt:
+excerpt: Running DC/OS on Google Compute Engine
 title: Running DC/OS on Google Compute Engine
 navigationTitle: GCE
 menuWeight: 3
 ---
-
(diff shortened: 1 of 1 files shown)


pages/1.10/installing/oss/cloud/gce/index.md
```diff
@@ -172,42 +172,40 @@ terraform apply \
...
 
 ```bash
 terraform apply -var-file desired_cluster_profile.tfvars
 ```
 
-## Redeploy an existing Agent
+## Redeploy an existing agent
 
-If you want to redeploy a problematic agent (i.e., storage filled up, not responsive, etc.), you can tell terraform to redeploy during the next cycle.
+If you want to redeploy a problematic agent, you can tell Terraform to redeploy during the next cycle.
```
bernadinm: What do you think if this line matched line # 117?

PR #1109, opened 2018-06-07
DCOS_OSS-3554 Correcting .tmpl file with incorrect code at transport_…

SarahStegall:
…encryption json file.

## Description
https://jira.mesosphere.com/browse/DCOS_OSS-3554

## Urgency
- [ ] Blocker 
- [ ] High
- [x] Medium

Edited 

services/include/security-configure-transport-encryption.tmpl

to add

"security": {
"transport_encryption": {
"enabled": <true|false default false>,

 

This will change all versions that call this template:

2.3.0-1.0.0

2.2.0-1.0.0

2.1.0-1.0.0

### DIFF
--- a/pages/services/include/security-configure-transport-encryption.tmpl
+++ b/pages/services/include/security-configure-transport-encryption.tmpl
@@ -26,7 +26,7 @@ Install the DC/OS {{ model.techName }} service including the following options i
         "service_account_secret": "<full path of service secret>",
         "security": {
             "transport_encryption": {
-                "enabled": true{{ model.security.plaintext }}
+                "enabled": <true|false default false>,
             }
         }
     }{{ model.security.extras }}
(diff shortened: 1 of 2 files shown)


pages/services/include/security-configure-transport-encryption.tmpl
```diff
@@ -26,7 +26,7 @@ Install the DC/OS {{ model.techName }} service including the following options i
         "service_account_secret": "<full path of service secret>",
         "security": {
             "transport_encryption": {
-                "enabled": true{{ model.security.plaintext }}
```
benclarkwood: Some services support plaintext (which is why this template was there). 

What's the issue that was occurring?

benclarkwood: Ideally we'd be able to just correct the template such that the formatting works out right.

@nickbp @elezar have either of you had better luck with special symbols in a template?

nickbp: Slathering it with triple brackets may help: `{{{foo}}}`

SarahStegall:
This template does not need to be corrected. I am deleting the PR and will fix this issue another way. Closing without merge.

PR #1173, opened 2018-07-02
Fixed checkpointed resources removal omission.

tillt:
## Description
Documentation on mount disk resources is outdated and incomplete.

When the user changes the agent resources, there is one step missing in the guide; removing the agent's checkpointed resources

https://jira.mesosphere.com/browse/DCOS_OSS-3713

## Urgency
- [x] Blocker 
- [ ] High
- [ ] Medium

## Requirements

### DIFF
--- a/pages/1.11/storage/mount-disk-resources/index.md
+++ b/pages/1.11/storage/mount-disk-resources/index.md
@@ -58,6 +58,12 @@ In this example, a disk resource is added to a DC/OS agent post-install on a run
         sudo rm -f /var/lib/mesos/slave/meta/slaves/latest
         ```
 
+    1. Remove checkpointed agent resources with this command:
+
+      ```bash
+      sudo rm -f /var/lib/mesos/slave/meta/resources
+      ```
+
 5. Create a 200 MB loopback device.
 
     ```bash
--- a/pages/1.11/storage/mount-disk-resources/index.md
+++ b/pages/1.11/storage/mount-disk-resources/index.md
@@ -17,7 +17,9 @@ When a DC/OS agent starts, it scans for volumes that match the pattern `/dcos/vo
 
 # Example using loopback device
 
-In this example, a disk resource is added to a DC/OS agent post-install on a running cluster. These same steps can be used pre-install without having to stop services or clear the agent state.
+In this example, a disk resource is added to a DC/OS agent post-install on a running cluster. These same steps can be used pre-install without having to stop services or clear the agent state. 
+
+Please note that this example handles **adding** resources exclusively and can not get applied the same way when removing resources.
 
 ***Warning:*** This will terminate any running tasks or services on the node.
 
@@ -57,12 +59,6 @@ In this example, a disk resource is added to a DC/OS agent post-install on a run
         ```bash
         sudo rm -f /var/lib/mesos/slave/meta/slaves/latest
         ```
-
-    1. Remove checkpointed agent resources with this command:
-
-      ```bash
-      sudo rm -f /var/lib/mesos/slave/meta/resources
-      ```
 
 5. Create a 200 MB loopback device.
 


pages/1.11/storage/mount-disk-resources/index.md
```diff
@@ -58,6 +58,12 @@ In this example, a disk resource is added to a DC/OS agent post-install on a run
         sudo rm -f /var/lib/mesos/slave/meta/slaves/latest
         ```
 
+    1. Remove checkpointed agent resources with this command:
+
+      ```bash
+      sudo rm -f /var/lib/mesos/slave/meta/resources
```
vinodkone: LGTM. Can you ask @chhsia0  to take a look as well?

Also, do we need to improve the log that Mesos agent itself spits out on recovery failures, to keep up with this documentation?

tillt: See https://issues.apache.org/jira/browse/MESOS-9046

chhsia0: I don't think we should remove `/var/lib/mesos/slave/meta/resources` even in this example, because it would make the agent lose all persistent volumes. If `/dcos/volume0` did not exist in the very beginning, the agent should be able to recover without removing this file.

The full error message is
```
... disk[MOUNT:/dcos/volume1]:5850 ... does not contain disk[MOUNT:/dcos/volume1]:5851
```
Which is from https://github.com/apache/mesos/blob/1.5.x/src/common/resources_utils.cpp#L85,
and this indicates that originally the checkpointed resources already had `disk[MOUNT:/dcos/volume1]:5851` with either dynamic reservation or persistence, and the original one is slightly bigger than the new `/dcos/volume1` added through the `--resources` agent flag. To safely replace `/dcos/volume1`, one should first destroy the persistent volume on it and unreserve it before doing the replacement.

porridge: So I have to admit that the original instructions did work initially, when the agent did not contain a persistent volume.

OTOH we might have discovered a different bug here. I admit that I did not fully understand this message, and now that you are saying that `:5851` is the size, I must stress that I did not touch the volumes. So if it came back with a different size, something is very wrong. All I did was the following:

```
sudo systemctl stop dcos-mesos-slave.service
sudo rm -f /var/lib/dcos/mesos-resources
sudo rm -f /var/lib/mesos/slave/meta/slaves/latest
function makevolume() {
  num="$1"
  sudo mkdir -p /dcos/volume${num}
  sudo dd if=/dev/zero of=/root/volume${num}.img bs=1M count=6000
  sudo losetup /dev/loop0 /root/volume${num}.img
  sudo mkfs -t xfs /dev/loop0
  sudo losetup -d /dev/loop0
  echo "/root/volume${num}.img /dcos/volume${num} auto loop 0 2" | sudo tee -a /etc/fstab
}
makevolume 0
makevolume 1
sudo reboot
# wait for the agent to come up, this worked fine
# now, simulate an "uninstall and reinstall" of DC/OS on the node:
sudo systemctl stop dcos-mesos-slave.service
sudo rm -f /var/lib/dcos/mesos-resources
sudo rm -f /var/lib/mesos/slave/meta/slaves/latest
sudo reboot
# at this point after booting the agent was stuck
```

porridge: Hm, I cannot reproduce this change in size now, using `master` DC/OS :-/
Agent comes up fine after both reboots. I tried storing a couple of megabytes on the volume to see if usage changes anything but no, still comes up fine.

SarahStegall: So what is your proposed resolution?

porridge: I think we should leave the document as is.
It looks like my problem was *not* the fact that `.../slave/meta/resources` existed, but the fact that the volume has mysteriously changed its size by one MB: `... disk[MOUNT:/dcos/volume1]:5850 ... does not contain disk[MOUNT:/dcos/volume1]:5851`. If we were to change anything, I think it's make this long message easier to parse for the un-initiated.
But I'd close this PR without merging.

tillt:
@SarahStegall your edits were unrelated to my changes - created a new branch and PR for your convenience: https://github.com/mesosphere/dcos-docs-site/pull/1220

PR #1186, opened 2018-07-03
DCOS-25686 Corrected file name in 1.9, 1.10 and 1.11 versions.

SarahStegall:
## Description
https://jira.mesosphere.com/browse/DCOS-25686
"Please note that the DC/OS provided as a service by Azure is deployed on Ubuntu servers, where the journalctl is located under /bin/journalctl and not /usr/bin/journalctl"

## Urgency
- [ ] Blocker 
- [ ] High
- [x] Medium

Versions fixed:

- [x] 1.09
- [x] 1.10
- [x] 1.11

### DIFF
--- a/pages/1.10/monitoring/logging/aggregating/elk/index.md
+++ b/pages/1.10/monitoring/logging/aggregating/elk/index.md
@@ -82,7 +82,7 @@ For all nodes in your DC/OS cluster:
     [Service]
     Restart=always
     RestartSec=5
-    ExecStart=/bin/sh -c '/usr/bin/journalctl --since="5 minutes ago" --no-tail --follow --unit="dcos*.service" >> /var/log/dcos/dcos.log 2>&1'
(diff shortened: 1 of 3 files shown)


leehambley:
The change looks good to me.

I wondered if it was better/plausible to rely on the setting of `Environment=` to set `PATH` to something that contains both paths so we don't have to special-case the Azure instance docs, but I'm not sure what, if any side-effects it would have (I think none, but I'm not sure):

```
sudo tee /etc/systemd/system/dcos-journalctl-filebeat.service<<-EOF
[Unit]
Description=DCOS journalctl parser to filebeat
Wants=filebeat.service
After=filebeat.service

[Service]
Restart=always
RestartSec=5
Environment="PATH=/bin:/usr/bin"
ExecStart=/bin/sh -c 'journalctl  --since="5 minutes ago" --no-tail --follow --unit="dcos*.service" >> /var/log/dcos/dcos.log 2>&1'

[Install]
WantedBy=multi-user.target
EOF
```

If this works, I think it might be more homogeneous, but I'm OK with the change as described so far.

PR #1196, opened 2018-07-06
Updated index.md to add correct proxy syntax

edgarlanting:
Since having a couple of our customers configure `https_proxy` with a `https` address and ending up having serious issues deploying their UCR based applications I'm adding this snippet to the guide to make this more clear.

With a proxy address starting with the HTTP scheme curl is still able to send HTTPS requests through the proxy – but it expects to have the connection between curl and the proxy not to be encrypted. That is sometimes called "HTTPS wrapped in plaintext", because the initial TCP connection between curl and the proxy appliance is not encrypted (and then, later, a TLS connection is established within that plaintext transport).

This explicitly instructs curl to use the HTTP proxy (not HTTPS proxy) for both, HTTP and HTTPS requests. In other words, this instructs curl to not even attempt to perform a TLS handshake when connecting to the proxy appliance.

What basically happens is that the system `curl` supports `https` and the DC/OS `curl` (e.g. the `curl` that you also use inside of `dcos-shell` does not). Please contact support if you need a link to the internal ticket that addresses this behaviour.

Our DC/OS supplied `curl` attempts to perform a handshake with their appliance before sending the `CONNECT`, and fails.

• System has curl 7.29 with NSS, DC/OS curl is 7.59 with OpenSSL.
• curl 7.29 is from 2013
• curl 7.52 (2016) added support for HTTPS proxies: proxy: Support HTTPS proxy and SOCKS+HTTP(s) (from the changelog: https://curl.haxx.se/changes.html)
this explains why the new curl actually respects the httpS in the proxy URL configuration and does not send a CONNECT over non-TLS

This commit added HTTPS proxy support to curl: https://github.com/curl/curl/commit/cb4e2be

From the commit message:

```
HTTPS proxies protect nearly all aspects of user-proxy communications as opposed to HTTP proxies that receive all requests (including CONNECT requests) in vulnerable clear text.
```

## Urgency
- [ ] Blocker 
- [x] High
- [ ] Medium

## Requirements

### DIFF
--- a/pages/1.11/installing/ent/custom/configure-proxy/index.md
+++ b/pages/1.11/installing/ent/custom/configure-proxy/index.md
@@ -16,5 +16,32 @@ When you configure DC/OS to configure a proxy during the installation process by
 
 ```
 http_proxy: http://<proxy-address>:<proxy-port>
+https_proxy: https://<proxy-address>:<proxy-port>
+```
+
+If this is however problematic and you cannot deploy any UCR based applications, please check the `stderr` log of the application you're trying to provision (e.g. Elastic) and see if the following error is occurring:
+
+```
+E0704 HH:MM:SS.MS 18419 fetcher.cpp:600] EXIT with status 1: Failed to fetch 'https://downloads.mesosphere.com/<package-name>/assets/<version>/<filename>.zip': Error downloading resource: SSL connect error
+Failed to synchronize with agent (it's probably exited)
+```
+
+If this is the case you can run the following test to see if you will need to make some changes.
+
+**1.)** Log into a cluster node and run `dcos-shell`.
+**2.)** Run a `curl -v <download-file-that-is-failing>.zip`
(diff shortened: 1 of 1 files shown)


edgarlanting:
Thats indeed correct @jgehrcke and the change indeed needs some additional information which I will add regarding to the curl test(s). My reasoning behind this is also because we've had 3 or 4 customers with the same "problem" now, hence my suggestion to the docs change.


What I'll do is change/replace those PR's to add the curl test procedure so customers can test before they go live. Thanks for the heads up!

edgarlanting:
Thats indeed correct @jgehrcke and the change indeed needs some additional information which I will add regarding to the curl test(s). My reasoning behind this is also because we've had 3 or 4 customers with the same "problem" now, hence my suggestion to the docs change.

What I'll do is change/replace those PR's to add the curl test procedure so customers can test before they go live. Thanks for the heads up!

edgarlanting:
Closing these PR's and doing some rewriting to clean up my suggestions.

PR #1261, opened 2018-07-30
Fix the table formatting for dcos-1.11.x cli

leehambley:
Because of blank lines and indentation the table was not rendering correctly. Confirmed in preview mode that this works. The only change is formatting of table markup. I looked at using markdown table syntax but it becomes absolutely unreadable.

No JIRA issue, but I was looking at `dcos tunnel` as part of triaging https://jira.mesosphere.com/browse/DCOS_OSS-2143

I haven't backported this change to older versions. After review let me know if I should.


leehambley:
Done, also made the same changes for 1.9, 1.10 and the new 1.12 branch after rebasing my fix over `staging`.

PR #1281, opened 2018-08-06
ss/DCOS-25881 Correcting package install section for /oss/ and /ent/ …

SarahStegall:
…versions of 1.8 only.

## Description
https://jira.mesosphere.com/browse/DCOS-25881

The last step in this page currently reads like this:

Perform steps from 2 to 7 of Installing the default Universe packages section exept of the 6th one. Run the following command instead:

$ dcos package repo add local-universe http://master.mesos:8082/repo

But it is out-dated. It should be modified to this:

Perform all of the steps in the Installing the default Universe packages section. But instead of using the default local-package container, use the package file you just built.

## Urgency
- [ ] Blocker 
- [ ] High
- [x] Medium

Corrections made to /oss/ and /ent/ versions of 1.8 only.


SarahStegall:
Since Cian O'Mahony is not available as a reviewer, I have assigned this to Marco, since he recently commented on it. If that's not appropriate, please assign it to a more appropriate reviewer.

PR #1318, opened 2018-08-10
Fixed generate node command in Patching docs.

anithamuthu79:
https://jira.mesosphere.com/browse/COPS-3668

## Urgency
- [ ] Blocker 
- [ ] High
- [ ] Medium

## Requirements


jrx:
@anithamuthu79 

Hi, I think still the commands that actually perform the patch/update are left out/invalid.
In both sections Masters and Agents, the correct command would be:

```
sudo bash dcos_node_upgrade.sh
```

instead of:

```
sudo bash dcos_node_patch.sh
```

Also this probably should also be changed for DC/OS 1.12.

Best,
Jan

PR #1365, opened 2018-08-21
1.11 - Updated volume mode field.

anithamuthu79:
https://jira.mesosphere.com/browse/DCOS-40228

## Urgency
- [ ] Blocker 
- [ ] High
- [ ] Medium

## Requirements

### DIFF
--- a/pages/1.12/storage/persistent-volume/index.md
+++ b/pages/1.12/storage/persistent-volume/index.md
@@ -451,7 +451,7 @@ The following example will create a pod with 2 containers and one shared persist
 
 ## Inspect and delete suspended stateful tasks
 
-Perform the following steps to free the reserved resources associated with a task, destroy and clean up persistent volumes. 
+To destroy and clean up persistent volumes and free the reserved resources associated with a task, perform the following two steps:
 
 1. Locate the agent containing the persistent volume and remove the data inside it.
 1. Send an HTTP DELETE request to Marathon that includes the `wipe=true` flag.
@@ -479,8 +479,9 @@ response:
 }
 ```
 
-**Note:** 
-A running task will show `stagedAt`, `startedAt`, and `version` in addition to the information provided above. You can do the following actions:
+**Note:** A running task will show `stagedAt`, `startedAt`, and `version` in addition to the information provided above.
+
+You can then
 
 1. Remove the data on disk by `ssh'ing` into the agent and running the `rm -rf <volume-path>/*` command.
 1. Delete the task with `wipe=true`, which will expunge the task information from the Marathon internal repository and eventually destroy the volume and unreserve the resources previously associated with the task:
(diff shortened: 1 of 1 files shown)


anithamuthu79:
@qzhang  could you review the changes made in 1.11? If these changes look good to you, I will make the same change in previous releases. Ty.

pages/1.12/storage/persistent-volume/index.md
```diff
@@ -38,7 +38,7 @@ Configure a persistent volume with the following options:
 ```
 
 - `containerPath`: The path where your application will read and write data. This must be a single-level path relative to the container; it cannot contain a forward slash (`/`). (`"data"`, but not `"/data"`, `"/var/data"` or `"var/data"`).
-- `mode`: The access mode of the volume. Currently, `"RW"` is the only possible value and will let your application read from and write to the volume.
+- `mode`: The access mode of the volume. Currently, `"RW"` and `"RO"` are the only possible values and will let your application read from and write to the volume.
```
qianzhangxa: "RW" will let your application read from and write to the volume, and "RO" will only let your application read from the volume.

qianzhangxa:
@anithamuthu79 Besides the changes about the volume mode field, I see you did some other changes which is not related to volume mode. Since the subject of this PR is `1.11 - Updated volume mode field.`, should those irrelevant changes be moved to a separate PR?

And I verified the `RW` and `RO` with Mesos but not with DC/OS, have you verified them in a DC/OS cluster from E2E? I guess we need to do that before updating the doc.

anithamuthu79:
@qianzhangxa who is the engineer to verify your statement?
"And I verified the RW and RO with Mesos but not with DC/OS, have you verified them in a DC/OS cluster from E2E? I guess we need to do that before updating the doc."

pages/1.12/storage/persistent-volume/index.md
```diff
@@ -451,7 +451,7 @@ The following example will create a pod with 2 containers and one shared persist
 
 ## Inspect and delete suspended stateful tasks
 
-To destroy and clean up persistent volumes and free the reserved resources associated with a task, perform 2 steps:
+Perform the following steps to free the reserved resources associated with a task, destroy and clean up persistent volumes.
```
qianzhangxa: The right order is, destroy and clean up persistent volumes first, and then free the reserved resources associated with a task. So I think the original content is correct in this regard.

pages/1.12/storage/persistent-volume/index.md
```diff
@@ -479,9 +479,8 @@ response:
 }
 ```
 
-**Note:** A running task will show `stagedAt`, `startedAt`, and `version` in addition to the information provided above.
-
-You can then
+**Note:** 
+A running task will show `stagedAt`, `startedAt`, and `version` in addition to the information provided above. You can do the following actions:
```
qianzhangxa: I do not think `You can do the following actions:` should be part of the `Note`.

pages/1.12/storage/persistent-volume/index.md
```diff
@@ -494,6 +493,6 @@ http DELETE http://dcos/service/marathon/v2/apps/postgres/tasks/postgres.53ab873
 
 After you have created your application, click the **Volumes** tab of the application detail view to get detailed information about your app instances and associated volumes.
 
-The Status column tells you if your app instance is attached to the volume or not. The app instance will read as "detached" if you have scaled down your application. Currently the only Operation Type available is read/write (RW).
+The `Status` column tells you if your app instance is attached to the volume or not. The app instance will read as "detached" if you have scaled down your application. Currently, the only operation type available are read/write (RW) and read only (RO). "RW" will let your application read from and write to the volume, and "RO" will only let your application read from the volume.
```
qianzhangxa: `operation type` should be `operation types`?

anithamuthu79:
@qianzhangxa could you review the changes on this PR?

qianzhangxa:
LGTM!

anithamuthu79:
@qianzhangxa can you approve this PR?

PR #1386, opened 2018-08-27
Updated Mount Disk Resources instructions to use cache system.

ArmandGrillet:
We now recommend to move the previous Mesos resources file to a cache
instead of removing it. Following the DC/OS update for 1.10/1.11/1.12,
this cache will automatically be used when creating the new resources
file and thus fix the issue described in DCOS_OSS-3921.

## Description
https://jira.mesosphere.com/browse/DCOS_OSS-3921

## Urgency
- [ ] Blocker 
- [ ] High
- [X] Medium

## Requirements

### DIFF
--- a/pages/1.10/storage/mount-disk-resources/index.md
+++ b/pages/1.10/storage/mount-disk-resources/index.md
@@ -46,10 +46,10 @@ In this example, a disk resource is added to a DC/OS agent post-install on a run
       ```
 
 4.  Clear agent state.
-    1. Remove volume mount discovery resource state with this command:
+    1. Cache volume mount discovery resource state with this command:
 
         ```bash
-        sudo rm -f /var/lib/dcos/mesos-resources
+        sudo mv -f /var/lib/dcos/mesos-resources /var/lib/dcos/mesos-resources.cache
         ```
 
     1. Remove agent checkpoint state with this command:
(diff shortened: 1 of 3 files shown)


pages/1.10/storage/mount-disk-resources/index.md
```diff
@@ -46,10 +46,10 @@ In this example, a disk resource is added to a DC/OS agent post-install on a run
       ```
 
 4.  Clear agent state.
-    1. Remove volume mount discovery resource state with this command:
+    1. Cache volume mount discovery resource state with this command:
```
adyatlov: Thanks for the contributing to the documentation. Do you think we can add a short sentence which explains why we need to cache this file? For example:
> Cache volume mount discovery resource state with the following command. DC/OS will check this file to generate a new resource configuration for the Mesos agent.

ArmandGrillet: I have added one line in `pages/1.12/storage/mount-disk-resources/index.md`. I have not backported it because the fix in DCOS_OSS-3921 has been backported to 1.10 and 1.11 but does not concern all 1.10 and 1.11 versions (but it will be in DC/OS 1.12.0).

adyatlov: Hi, @ArmandGrillet. It makes sense to me. Please, ship it.

PR #1538, opened 2018-09-27
ss/COPS-3618 Deleted --strict from Create-a-secret bash file.

SarahStegall:
## Description
https://jira.mesosphere.com/browse/COPS-3618

I followed the documentation to install EdgeLB on a disabled security cluster and the EdgeLB did install successfully. But I was unable to add a pool.

On further investigation (COPS-3603) I discovered that the problem was here https://docs.mesosphere.com/services/edge-lb/1.0/installing/#create-a-secret

For disabled security cluster

dcos security secrets create-sa-secret edge-lb-private-key.pem edge-lb-principal dcos-edgelb/edge-lb-secret
should be used instead. Notice the --strict removed

## Urgency
- [ ] Blocker 
- [x] High
- [ ] Medium

Deleted "--strict" from bash file as requested. 
Versions changed:
       Edge-LB 1.0

Please let me know if this issue also affects Edge-LB 1.1 and 1.2


SarahStegall:
I added back the --strict parameter for non-disabled security modes. Please review. Also, should I make the same changes to versions 1.1 and 1.2?

surdy:
Yes, same applies to 1.2 as well

PR #1540, opened 2018-09-28
Document intra-cluster network access explicitly

jongiddy:
## Description
https://jira.mesosphere.com/browse/DCOS-42708

Document the network rules required inside the cluster. Provides an explicit list of all the ports to be allowed or blocked between every type of node.

The list is removed from the ports page, to reduce duplication, but is linked from that page.

## Urgency
- [ ] Blocker 
- [x] High
- [ ] Medium

## Requirements

### DIFF
--- a/pages/1.10/administering-clusters/securing-your-cluster/index.md
+++ b/pages/1.10/administering-clusters/securing-your-cluster/index.md
@@ -27,12 +27,14 @@ Depending on your cluster environment, this may include:
 - using router firewalls to restrict access to ports;
 - using firewall software (e.g. `iptables`) on the nodes to restrict access to ports.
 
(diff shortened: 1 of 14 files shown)


jongiddy:
@jrx I want these instructions to be simple and mechanistic, since they may get passed to a different person to implement. So, I don't want to explain them on the network security page. The right thing is to add them to the ports list, since it describes ports that need to be available and accessible for DC/OS to work.

jrx:
@jongiddy I'm fine having the ports on the ports list and pointing to it, just wanted to point out that 8201 and 26257 are currently missing on the ports list. :)

jgehrcke:
Thank you for this!

CC @somikbehera

PR #1584, opened 2018-10-11
Update default containerizer type in "security/ent/index.md".

rukletsov:
Looking at defaults in 1.11 strict and 1.12-rc2 strict,
I see this is not the case: the default containerizer
is Docker, not Mesos (UCR).

Also clarify that Mesos is the same as UCR.

## Urgency
- [ ] Blocker 
- [ ] High
- [x] Medium

## Requirements

### DIFF
--- a/pages/1.10/security/ent/index.md
+++ b/pages/1.10/security/ent/index.md
@@ -233,13 +233,13 @@ DC/OS allows you to restrict:
 
 The default Linux user for tasks and sandbox files varies according to your [security mode](/1.10/security/ent/#security-modes) and the [type of container](/1.10/deploying-services/containerizers/) the task runs inside of.
 
-By default, all tasks will run inside of Mesos containers. However, a user service can be configured to run tasks inside of Docker containers instead. Please see [Deploying a Docker-based Service to Marathon](/1.10/deploying-services/creating-services/deploy-docker-app/) for an example.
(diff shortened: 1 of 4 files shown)


pages/1.10/security/ent/index.md
```diff
@@ -233,13 +233,13 @@ DC/OS allows you to restrict:
 
 The default Linux user for tasks and sandbox files varies according to your [security mode](/1.10/security/ent/#security-modes) and the [type of container](/1.10/deploying-services/containerizers/) the task runs inside of.
 
-By default, all tasks will run inside of Mesos containers. However, a user service can be configured to run tasks inside of Docker containers instead. Please see [Deploying a Docker-based Service to Marathon](/1.10/deploying-services/creating-services/deploy-docker-app/) for an example.
+By default, all tasks will run inside of Docker containers. However, a user service can be configured to run tasks inside of Docker containers instead. Please see [Deploying a Docker-based Service to Marathon](/1.10/deploying-services/creating-services/deploy-docker-app/) for an example.
```
pavisandhu: With this change, the 2nd sentence no longer makes sense, so should be deleted. Unless I'm misunderstanding something, the para should say: 
"By default, all tasks will run inside of Docker containers. Please see [Deploying a Docker-based Service to Marathon](/1.10/deploying-services/creating-services/deploy-docker-app/) for an example."
The same change would apply to all 4 versions.

ck4adventure: I'm merging this and then turning around and updating the sentence on a new PR to get it done quickly

tillt: Thanks for the fixup and sorry I missed this!

PR #1650, opened 2018-11-02
ss/COPS-3970 Removing all $ from beginning of bash command lines.

SarahStegall:
## Description
https://jira.mesosphere.com/browse/COPS-3970
Throughout the documentation, there are many code blocks that can be 'copied. An example is https://docs.mesosphere.com/services/cassandra/2.3.0-3.0.16/getting-started/#install-a-basic-cluster where the line reads

$ dcos package install cassandra
.

However, when clicking on the copy button to then enter the code into a shell, it also copies the dollar sign at the start of the line. This defeats the purpose of the copy button because the user has to then go back and edit the line to remove the dollar sign.

It becomes worse when there are multiple lines of code in each block, because the user then has to manually copy each line one by one.

This ticket is to request that throughout the documentation, ALL copyable code blocks have the dollar sign removed.
## Urgency
- [ ] Blocker 
- [ ] High
- [x] Medium

303 files changed to remove $ at beginning of bash command lines.


pages/cn/services/Spark-2-3-1-2-2-1-2/job-scheduling/index.md
```diff
@@ -83,7 +83,7 @@ $ cat options.json
         "role": "dispatcher"
     }
 }
-$ dcos package install spark --options=options.json
+dcos package install spark --options=options.json
```
niartenyaw: So not sure what we want to do about cases like this. Here, the `$` distinguishes a command that should be entered and everything else is the output of that command.

I think for these cases, the command and output should just be split up between multiple code blocks so the copy to clipboard functionality works correctly (to both remove `$` and not copy the output to the clipboard). Not sure when we want to do this, but we would need to look at every code block in our docs 😬

SarahStegall: *head desk*

SarahStegall:
Since there is so much ambiguity regarding when and where the $ appears in a code block, we need a much larger effort than is reflected here. I am closing this Pull Request and putting the JIRA ticket on a back burner until we decide on a process.

PR #1670, opened 2018-11-07
ss/DCOS-44578 Re-order compatibility matrix, add Oracle.

SarahStegall:
## Description
https://jira.mesosphere.com/browse/DCOS-44578

We have some updates needed for the Version Policy page.
https://docs.mesosphere.com/version-policy/

Specifically, with the DC/OS Platform Version Compatibility Matrix:

1. List the OSes by recency (i.e CentOS 7.5, 7.4, 7.3, CoreOS 1855, then CoreOS 1800.7, etc...) rather than from the opposite order (which is currently the oldest version of the OS first). THis is because we want to push users to the latest & greatest and show our support for that.

2. Please add "Oracle Linux 7.5 (RHCK), Oracle Linux 7.4 (RHCK), Oracle Linux 7.3 (RHCK)" as Platform component that is supported on all 3 latest stable of "DC/OS 1.12, 1.11, 1.10" - the bullet point should be the same as the CentOS/RHEL of "RH Fork of Docker CE 1.13.1".

## Urgency
- [ ] Blocker 
- [x] High
- [ ] Medium

1. Components re-ordered by age, as requested.
2. Oracle components added and ranked by number.

### DIFF
--- a/pages/version-policy/index.md
+++ b/pages/version-policy/index.md
@@ -3,14 +3,14 @@ layout: layout.pug
 navigationTitle:  Version Policy
 title: Version Policy
 menuWeight: 5
-excerpt:
+excerpt: DC/OS version lifecycle and compatibility matrix
 ---
 
-This page was updated on October 23, 2018. 
+This page was updated on November 7, 2018. 
 
 # Mesosphere DC/OS Version Lifecycle and Compatibility Matrix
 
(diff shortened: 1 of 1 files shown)


yishan-lin:
Looks great - thank you Sarah!

PR #1735, opened 2018-11-21
ck/COPS-4158 Update Haproxy_Head script in example

ck4adventure:
## Description
[https://jira.mesosphere.com/browse/COPS-4158](https://jira.mesosphere.com/browse/COPS-4158)

## Urgency
- [ ] Blocker 
- [X ] High
- [ ] Medium

## Requirements

### DIFF
--- a/pages/services/marathon-lb/advanced/index.md
+++ b/pages/services/marathon-lb/advanced/index.md
@@ -104,7 +104,7 @@ To specify a global template:
     tar czf templates.tgz templates/
     ```
 
-    Take the file you created (`templates.tgz` if you use the script), and make it available from an HTTP server. 
+    Take the file you created (`templates.tgz`), and make it available from an HTTP server. 
 
 3.  Augment the Marathon-LB config by copying the following JSON into a file called `options.json`, putting in the URL to your file:
 
--- a/pages/services/marathon-lb/advanced/index.md
+++ b/pages/services/marathon-lb/advanced/index.md
@@ -98,15 +98,19 @@ To specify a global template:
 
     The current `HAPROXY_HEAD`, as well as other Marathon templates, can be [found here](https://github.com/mesosphere/marathon-lb/blob/master/Longhelp.md#haproxy_head").
 
-2.  Tar or zip the file. [Here is a handy script you can use to do this][1].
+2.  Tar or zip the file. 
 
-Take the file you created (`templates.tgz` if you use the script), and make it available from an HTTP server. If you would like to use the sample one, use this URI: <https://downloads.mesosphere.com/marathon/marathon-lb/templates.tgz>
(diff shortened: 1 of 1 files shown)


pages/services/marathon-lb/advanced/index.md
```diff
@@ -31,38 +31,72 @@ Marathon-LB has a templating feature for specifying custom HAProxy configuration
...
+      http-request use-service lua.signalmlbhup if signalmlbhup
+      acl signalmlbusr1 path /_mlb_signal/usr1
+      http-request use-service lua.signalmlbusr1 if signalmlbusr1
+    ```
+
+    In the code above, the following items have changed from the default: `maxconn`.
+
+    The current `HAPROXY_HEAD`, as well as other Marathon templates, can be [found here](https://github.com/mesosphere/marathon-lb/blob/master/Longhelp.md#haproxy_head").
 
 2.  Tar or zip the file. [Here is a handy script you can use to do this][1].
```
cneth: Sorry, I just noticed this - not your fault,  but I think this "handy script" thing is problematic, for the following reasons:

1) it contains another copy of the old, outdated version of the template.
2) It points at an ex-employee's personal github acccount; i.e. they could delete it any time.  Also, we can't change it (we'd have to make a new one and host it somewhere)
3) It is not really all that handy, it's basic shell script stuff.  It just makes this page more complicated to maintain and follow without really helping the end-user.

I'd like to propose we just change this line to:
---------------
2.  Tar up the file: `tar czf templates.tgz templates/`
---------------
Note that the next line in this file references this script "if you use the script", I'd just delete that reference.

pages/services/marathon-lb/advanced/index.md
```diff
@@ -31,48 +31,86 @@ Marathon-LB has a templating feature for specifying custom HAProxy configuration
...
+
+2.  Tar or zip the file. 
+
+    ```bash
+    tar czf templates.tgz templates/
+    ```
+
+    Take the file you created (`templates.tgz` if you use the script), and make it available from an HTTP server. 
+
+3.  Augment the Marathon-LB config by copying the following JSON into a file called `options.json`, putting in the URL to your file:
```
ck4adventure: So I looked at the step below that one.... and saw that it points the template-url to that old download again. So I am editing that line too, to instruct the user to put the URL of their own file in the options.json

cneth: Yes, this looks good.  My only other comment is on line 107, we should change:
(`templates.tgz` if you use the script)   --> (`templates.tgz`)
Because, we got rid of the 'the script'.

ck4adventure: done :)

PR #1803, opened 2018-12-11
ss/DCOS-45629 Editing Edge-LB uninstall instructions for DC/OS 1.10.

SarahStegall:
## Description
https://jira.mesosphere.com/browse/DCOS-45629
## Urgency
- [ ] Blocker 
- [x] High
- [ ] Medium
Under “Known Limitation”, I see “The steps provided in the DC/OS web interface to uninstall Edge-LB are incorrect. Follow these steps in the [Edge-LB uninstall documentation](/services/edge-lb/1.2/uninstalling/).”

Do you think you could change it to:

The steps provided in the DC/OS 1.10 web interface to uninstall Edge-LB are incorrect. In order to correctly uninstall Edge-LB for any given DC/OS version, please follow the steps in the Edge-LB uninstall documentation. (the link of course stays :slightly_smiling_face: ). 
The problem is that the warning message is broken only for 1.10. The 1.11+ does not have any message, but still - the uninstall instructions should be followed. As the Edge-LB cannot be simply uninstalled via the UI.

The changes need to be done for release notes for all Edge-LB versions.

### DIFF
--- a/pages/1.10/deploying-services/uninstall/index.md
+++ b/pages/1.10/deploying-services/uninstall/index.md
@@ -8,9 +8,8 @@ excerpt: Uninstalling DC/OS services from the CLI
 enterprise: false
 ---
 
+Most services can be uninstalled from the CLI or the DC/OS web interface. If a Universe service has any reserved resources that could not be be cleaned up by the normal uninstall process, you may also need to run the framework cleaner script. The [framework cleaner script](#framework-cleaner) removes the service instance from ZooKeeper, along with any data associated with it.
 
-
-Services can be uninstalled from the CLI. If a Universe service has any reserved resources that could not be be cleaned up by the normal uninstall process, you may also need to run the framework cleaner script. The [framework cleaner script](#framework-cleaner) removes the service instance from ZooKeeper, along with any data associated with it.
 
 # Uninstalling Universe services
 
@@ -30,6 +29,8 @@ dcos package uninstall chronos
 
 ## Web interface
 
+<p class="message--note"><strong>NOTE: </strong>The Edge-LB service cannot be uninstalled using this procedure. Please follow the steps in the <a href="/services/edge-lb/1.2/uninstalling/">Edge-LB uninstall documentation</a>.</p>
+
 You can uninstall services from the DC/OS web interface from the **Services** tab. The Services tab provides a full-featured interface to the native DC/OS Marathon instance.
 
 1.  Navigate to the [**Services**](/1.10/gui/services/) tab in the DC/OS web interface.
@@ -41,7 +42,7 @@ You can uninstall services from the DC/OS web interface from the **Services** ta
 
 1.  Copy and run the displayed command.
 
-## Troubleshooting
+# Troubleshooting
 
 It is possible for an uninstall to fail with the following error message:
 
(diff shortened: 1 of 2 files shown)


SarahStegall:
You state that "the warning message is broken only for 1.10", and that the Edge-LB cannot be  uninstalled via the web  interface. I have added a note to the 1.10/deploying-services/uninstall page, under the Web interface uninstall instructions, to the effect that users should use the Edge-LB uninstall documentation. My question is: should I add a similar note to the uninstall pages of the subsequent releases?

pages/1.10/deploying-services/uninstall/index.md
```diff
@@ -8,10 +8,9 @@ excerpt: Uninstalling DC/OS services from the CLI
 enterprise: false
 ---
 
+Most services can be uninstalled from the CLI or the DC/OS web interface. If a Universe service has any reserved resources that could not be be cleaned up by the normal uninstall process, you may also need to run the framework cleaner script. The [framework cleaner script](#framework-cleaner) removes the service instance from ZooKeeper, along with any data associated with it.
```
vespian: As a user, how do I qualify `Most services can be uninstalled`? How can I say that the service I am trying to uninstall qualifies as the "most"?

pages/1.10/deploying-services/uninstall/index.md
```diff
@@ -30,6 +29,8 @@ dcos package uninstall chronos
 
 ## Web interface
 
+<p class="message--note"><strong>NOTE: </strong>The Edge-LB service cannot be uninstalled using this procedure. Please follow the steps in the <a href="/services/edge-lb/1.2/uninstalling/">Edge-LB uninstall documentation</a>.</p>
```
vespian: You are linking here to version 1.2, at some point we will have version 1.3, 1.4, etc... Shouldn't we mention here something like: `Please follow the steps outlined in the uninstall section of the edge-lb version documentation you are using`

pages/services/edge-lb/1.2/release-notes/index.md
```diff
@@ -35,7 +35,7 @@ Released on November 27, 2018.
 
 ## Known Issues
 
-* The steps provided in the DC/OS web interface to uninstall Edge-LB are incorrect. Follow these steps in the [Edge-LB uninstall documentation](/services/edge-lb/1.2/uninstalling/).
+* The steps provided in the DC/OS 1.10 web interface to uninstall Edge-LB are incorrect. In order to correctly uninstall Edge-LB for any given DC/OS version, please follow the steps in the [Edge-LB uninstall documentation](/services/edge-lb/1.2/uninstalling/).
```
vespian: See my comment about linking to particular edge-lb versions.

pages/1.10/deploying-services/uninstall/index.md
SarahStegall: By "most", I mean "not all". This covers those services which can only be uninstalled from the CLI, or services which can only be uninstalled from the web interface, as well as those which can be uninstalled by either method. I'm not sure I understand your second sentence. "The most" does not have the same meaning as "most"; "the most" is a superlative like "the greatest", but "most" just means "many" or "the majority". I am unsure which meaning you are referring to.

pages/1.10/deploying-services/uninstall/index.md
SarahStegall: When we release version 1.3, etc, it will have its own set of release notes, just as 1.1 and 1.0 have their own release notes. We do not combine release notes from different releases into one document.

pages/services/edge-lb/1.2/release-notes/index.md
SarahStegall: As noted, each release will have its own release notes, so this link is correct.

vespian:
@SarahStegall Why this PR was merged?

CC: @orlandohohmeier @GoelDeepak

PR #1807, opened 2018-12-11
ss/DCOS_OSS-3181 Editing HDFS pages.

SarahStegall:
## Description
https://jira.mesosphere.com/browse/DCOS_OSS-3181
## Urgency
- [ ] Blocker 
- [ ] High
- [x] Medium
DESCRIPTION
1. The Spark on HDFS chapter should be split into two chapters to avoid confusion.  The first chapter should cover the Spark on HDFS chapter and the second should cover SPARK on a Kerberorized HDFS cluster.  As presented it is confusing where one chapter ends and the Kerberos chapter begins.
2. Specifically, one more comment about the Spark on HDFS section.  For connecting to a DC/OS managed HDFS cluster where can a user find the dc/os dns URL and port numbers.  Just saying you should pull those values will leave many users scratching their head wondering what to do.  

EDITOR'S RESPONSE

1. The HDFS page is already divided between Spark on HDFS and Spark on HDFS Kerberos. However, I have added a section title to make the distinction clearer. Also edited some formatting issues.
2. It is not clear to me exactly what changes are requested here. Please clarify. Where WOULD users find the URL and port numbers?

VERSIONS AFFECTED:
1.0.9
1.1.0
1.1.1
2.0.0
2.0.1
2.1.0

### DIFF
--- a/pages/services/spark/2.1.0-2.2.0-1/hdfs/index.md
+++ b/pages/services/spark/2.1.0-2.2.0-1/hdfs/index.md
@@ -3,12 +3,12 @@ layout: layout.pug
 navigationTitle:  Configure Spark for HDFS
 title: Configure Spark for HDFS
 menuWeight: 20
-excerpt:
+excerpt: Configure Spark on HDFS and Kerberos
 featureMaturity:
 enterprise: false
 ---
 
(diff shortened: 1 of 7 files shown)


markfjohnson:
Looking at the docs, it looks like it was resolved.  You can close this out.

PR #1861, opened 2019-01-10
Added an example of json-encoded CSR.

anithamuthu79:
https://jira.mesosphere.com/browse/COPS-4349

## Urgency
- [ ] Blocker 
- [ ] High
- [ ] Medium

## Requirements


mesoshaun:
I don’t like the sentence:

An example of the content in csr file `<json-encoded-value-of-request.csr>` is repesented in a single line, where new lines are replaced with `\n`.

We've not giving an example of the content of a csr file, we're giving an example of the <json-encoded-value-of-request.csr> field.

'represented' is also mis-spelled.

I think the sentence could be rewritten as something like:

The `<json-encoded-value-of-request.csr>` field represents the content of the csr file as a single line, where new lines are replaced with `\n`.

PR #1909, opened 2019-01-24
ss/DCOS-26386 Added note on local interfaces to networking page.

SarahStegall:
## Description
https://jira.mesosphere.com/browse/DCOS-26386

Documentation page: https://docs.mesosphere.com/1.9/networking/#dns-forwarder-spartan
I cannot understand the following sentence in the DNS Forwarder (Spartan) section:

The Spartan instance on each agent is configured to listen to three different local interfaces on the agent and the nameservers on the agent are set to these three interfaces.

Also, that would be great to list these three interfaces.

UPD: here is the change which I think would make the sentence a bit more useful:

The Spartan instance on each agent is configured to listen to three different local interfaces: 198.51.100.1, 198.51.100.2, and 198.51.100.3. The nameserver option in the /etc/resolv.conf on the agent is set to these three interfaces.

## Urgency
- [ ] Blocker 
- [ ] High
- [x] Medium

Changed text as required in versions 1.9 and 1.10.


SarahStegall:
@adyatlov Yes. According to https://docs.mesosphere.com/1.11/networking/#a-note-on-software-re-architecture, "Prior to DC/OS 1.11, the role of dcos-dns was fulfilled by spartan, dcos-l4lb was fulfilled by minuteman and dcos-overlay was fulfilled by navstar". Spartan is deprecated in 1.11 and later versions.

PR #1910, opened 2019-01-25
ss/DCOS_OSS-3111 Adding disclaimer to Deploying a Local Universe pages.

SarahStegall:
## Description
https://jira.mesosphere.com/browse/DCOS_OSS-3111

> Source: https://docs.mesosphere.com/1.10/administering-clusters/deploying-a-local-dcos-universe
> 
> It doesn't appear that we support creating a local universe on MacOS. Can we note that explicitly on the prerequisites section? Prospect was struggling with doing this and we later determined that we don't actually support this.

## Urgency
- [ ] Blocker 
- [ ] High
- [ ] Medium

Added the following disclaimer to all affected versions in Prerequisites section:

> NOTE: Creation of a local universe on MacOS is not supported.

Versions affected:

- 1.9
- 1.10
- 1.11
- 1.12

Also removed "Preview" tag from 1.9 page.

### DIFF
--- a/pages/1.10/administering-clusters/deploying-a-local-dcos-universe/index.md
+++ b/pages/1.10/administering-clusters/deploying-a-local-dcos-universe/index.md
@@ -5,14 +5,14 @@ title: Deploying a Local Universe
 menuWeight: 1000
 excerpt:
 enterprise: false
-preview: true
+preview: false
 ---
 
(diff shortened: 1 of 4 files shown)


SarahStegall:
Josh Earlenbaugh concurs that this fix is appropriate and approved. Merging.

PR #2069, opened 2019-03-14
update index.md

c-mcinerney:
Add information regarding the format of the cluster_name that is the root for all other variables in the deployment.  Azure only allows lowercase alphanumeric characters, numbers and hyphens, and it must start with a letter and end only with a number or letter.

## Urgency
- [ ] Blocker 
- [ ] High
- [ ] Medium

## Requirements


ck4adventure:
We'll need to get that changed down around line 228 and for now 4 versions 1.10-1.13. If there isn't a jira ticket we'll need one.

ck4adventure:
This info is explained two steps later in the process. Please open a jira if there is a direct customer issue.

PR #2138, opened 2019-04-12
Update command and removed $

millerjem:
Updated the dcos repository addition command to use the dcos package repo add and removed all the $ which cause a problem when using the copy function.

## Urgency
- [ ] Blocker 
- [x] High
- [ ] Medium

## Requirements

### DIFF
--- a/pages/services/beta-storage/0.5.2-beta/install/package-registry-based/index.md
+++ b/pages/services/beta-storage/0.5.2-beta/install/package-registry-based/index.md
@@ -25,15 +25,15 @@ The `.dcos` package of the DC/OS Storage Service (DSS) can be downloaded from [M
 
 # Install the DSS package
 
-1. Assume that the downloaded package is called `beta-storage.dcos` in the current working directory. Run the `dcos registry add` command to add it.
+1. Assume that the downloaded package is called `beta-storage.dcos` in the current working directory. Run the `dcos package repo add` command to add it.
 
     ```bash
-    $ dcos registry add --dcos-file beta-storage.dcos
+    dcos package repo add Universe beta-storage.dcos
     ```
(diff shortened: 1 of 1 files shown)


bbannier:
The change of `registry` to `repo` looks good to me, thanks for fixing this.

For changes the `$` I already commented on in #2138,
>Is this a style we use consistently? On a global scale iIt is a slightly unusual style for code samples since it makes it hard to discover what is a command to be entered and what is echo'ed results. Would be great to codify this in some style guide if it isn't already documented.

> We use the same formatting trick for code samples in e.g., CLI commands of {{dcos storage}}.

> Could you please involve @jdef for review?

jdef:
I just commented on #2139  - we need to be consistent across the docs. The rest of the storage docs all use `$` notation for shell commands. If that's changed, then we need to do it everywhere vs. one-off PRs like this.

Can we revert this, please?

jpe442:
This PR must be made against the source of truth for these docs here: https://github.com/mesosphere/dcos-storage.

SarahStegall:
@bbannier  I have just updated the [style guide](https://wiki.mesosphere.com/display/DOCS/Documentation+Style+Guide) to explain _why_ we do not retain the shell prompts:

"Our code blocks are formatted with a copy icon in the upper right corner. Users can click on this icon and copy the entire code to their clipboard, rather than have to type all the code in. If the shell prompt is retained in the code, it will be copied and pasted into the user's shell prompt." Which we don't want to happen.

@jdef Deleting the shell prompts is an ongoing process, which we are implementing as we edit the docs. Right now, a wholesale purge is just not within our resources, so our choice is to delete the shell prompts as we find them during the edit process for another issue, or to leave them in place.

PR #2139, opened 2019-04-12
Remove the dollar-sign prompts

millerjem:
Remove the dollar-sign prompts which cause problems when using the copy functionality. I also separated the code blocks and updated the cat script to work when someone using the copy functionality.

## Urgency
- [ ] Blocker 
- [x] High
- [ ] Medium

## Requirements

### DIFF
--- a/pages/services/beta-storage/0.5.2-beta/install/index.md
+++ b/pages/services/beta-storage/0.5.2-beta/install/index.md
@@ -23,9 +23,9 @@ The DC/OS Storage Service may be run on DC/OS clusters in either permissive or s
 The following CLI commands create a service account named `storage-principal` and stores its private certificate in a secret named `storage/storage-private-key`:
 
 ```bash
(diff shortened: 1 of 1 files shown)


bbannier:
Is this a style we use consistently? On a global scale iIt is a slightly unusual style for code samples since it makes it hard to discover what is a command to be entered and what is echo'ed results. Would be great to codify this in some style guide if it isn't already documented.

We use the same formatting trick for code samples in e.g., CLI commands of {{dcos storage}}.

Could you please involve @jdef for review?

jdef:
Thanks for the feedback.

We use `$ dcos ...` notation everywhere for executing dcos storage commands. AFAIK this was the recommended approach. If that's changed, then it needs to be addressed a whole vs. one-off PRs like this.

jpe442:
@millerjem: This PR needs to be made against storage repo https://github.com/mesosphere/dcos-storage, as that is where the source of truth is for these docs. If we make the changes here, they will only be overwritten once storage repo makes its next update.

PR #2142, opened 2019-04-12
Revise content codeowners

ck4adventure:
## Description
https://jira.mesosphere.com/browse/DCOS-51783

Removing codeowners automatic notifications as it has been found to be confusing and not helping workflow and process.

## Urgency
- [ ] Blocker 
- [X ] High
- [ ] Medium

## Requirements

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,16 +1,21 @@
+# changes 4-12-19 CK
 # Aaron should see all code changes (things not in /pages)
-# taking off from ALL @jpe442
+# change to only notify Aaron here
 * @niartenyaw
 
 # Josh take over all content (remove Aaron)
+# change to not notify
 # /pages/ @jpe442
 
 # Sarah for services
+# change to not notify
 # /pages/services/ @SarahStegall @jpe442
 
 # Networking
-# /pages/services/edge-lb/ @networking-team @SarahStegall @jpe442
-# /pages/*/networking/ @networking-team @jpe442
+# change to notify networking team only 
+/pages/services/edge-lb/ @networking-team
+/pages/*/networking/ @networking-team
 
 # Storage
-# /pages/services/beta-storage/ @SarahStegall @jpe442 @mesosphere/mesosphere-storage
+# change to notify storage team only
+/pages/services/beta-storage/ @mesosphere/mesosphere-storage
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -1,16 +1,16 @@
 # Aaron should see all code changes (things not in /pages)
-# Josh should see ALL PRs
-* @niartenyaw @jpe442
+# taking off from ALL @jpe442
+* @niartenyaw
 
 # Josh take over all content (remove Aaron)
-/pages/ @jpe442
+# /pages/ @jpe442
```
niartenyaw: There still needs to be this line with no one to take me off

CODEOWNERS
```diff
@@ -1,16 +1,16 @@
...
 
 # Sarah for services
-/pages/services/ @SarahStegall @jpe442
+# /pages/services/ @SarahStegall @jpe442
 
 # Networking
-/pages/services/edge-lb/ @networking-team @SarahStegall @jpe442
-/pages/*/networking/ @networking-team @jpe442
+# /pages/services/edge-lb/ @networking-team @SarahStegall @jpe442
+# /pages/*/networking/ @networking-team @jpe442
```
niartenyaw: these should still be here, just without josh

CODEOWNERS
ck4adventure: You still get all, I left yours open on line 3, no??

niartenyaw: yeah, but i don't want to get things in `/pages`

CODEOWNERS
```diff
@@ -1,16 +1,21 @@
...
 
 # Josh take over all content (remove Aaron)
-/pages/ @jpe442
+# change to not notify
+# /pages/ @jpe442
 
 # Sarah for services
-/pages/services/ @SarahStegall @jpe442
+# change to not notify
+# /pages/services/ @SarahStegall @jpe442
```
niartenyaw: i would just delete this section

ck4adventure: whichever, I can see us maybe needing to ping /services in the future, but we can just delete it

CODEOWNERS
ck4adventure: What happens if I leave /pages blank?? Will it still cancel it out for you by order of precedence but not be able to assign anyone??

PR #2201, opened 2019-04-29
Revert "Ss/dcos 52140"

SarahStegall:
Reverts mesosphere/dcos-docs-site#2186


SarahStegall:
These edits should go into Chinese doc site.

PR #2210, opened 2019-05-01
1.13 Metrics landing page, updated content for Metrics architecture page, and added Telegraf diagram.

anithamuthu79:
https://jira.mesosphere.com/browse/DCOS-51694

## Urgency
- [ ] Blocker 
- [ ] High
- [ ] Medium

## Requirements


pages/1.13/metrics/index.md
```diff
@@ -10,37 +10,61 @@ enterprise: false
...
+# Operations on Metrics 
+You can perform the following operations on Metrics:
 
-Telegraf collects application and custom metrics through the `dcos_statsd` plugin. A dedicated StatsD server is started for each new task. Any metrics received by the StatsD server are tagged with the task name and its service name. The address of the server is provided by environment variables (`STATSD_UDP_HOST` and `STATSD_UDP_PORT`). Note that when a task finishes, any metrics it has emitted that haven't yet been gathered by Telegraf will be discarded. The metrics collected by `dcos_statsd` are gathered every 30 seconds. To ensure a task's metrics are gathered, the task must run for at least 30 seconds.
+1. **Adding Metrics:** You can add custom metrics to the DC/OS metrics service from your applications. DC/OS metrics listens for `StatsD` metrics from every application running with the Mesos containerizer. A `StatsD server` is exposed for each container, which allows you to tag all metrics by origin. Its address is available to the application by injecting the standard environment variables `STATSD_UDP_HOST` and `STATSD_UDP_PORT`.
```
branden: I think we should use "StatsD" without monospacing it, as it's used in StatsD docs, e.g. https://github.com/statsd/statsd/blob/master/docs/backend.md. StatsD isn't a command or string literal, so I don't think it's helpful to wrap it in backticks.

branden: @gracedo @philipnrmn We support collecting workload metrics in Prometheus format, right? Should we add something here about that?

philipnrmn: We should! Let's draft something here for that.

philipnrmn:
```suggestion
1. **Adding Metrics:** You can add custom metrics to the DC/OS metrics service from your applications. DC/OS metrics listens for `StatsD` metrics from every application running with the Mesos containerizer. A `StatsD server` is exposed for each container, which allows you to tag all metrics by origin. Its address is available to the application by injecting the standard environment variables `STATSD_UDP_HOST` and `STATSD_UDP_PORT`.

Alternatively, you can serve metrics in prometheus format by exposing an endpoint which serves metrics. Notify the metrics service that your task is serving prometheus metrics by adding a label to the port which is serving metrics. Any port with the label `DCOS_METRICS_FORMAT=prometheus` will be scraped at the rate of once per minute for the lifetime of the task. 

A task may expose multiple ports which serve prometheus metrics. Metrics collected in this manner will be tagged with the originating task's name and its framework name. 
```

philipnrmn: Here's a snippet of a task which exposes Prometheus format metrics:
```json
        "discovery": {
          "name": "task-with-metrics",
          "ports": {
            "ports": [
              { "number": 12345,
                "protocol": "tcp",
                "labels": {
                  "labels": [ {"key": "DCOS_METRICS_FORMAT","value": "prometheus"} ]
                }
              }
            ]
          }
        },
```

pages/1.13/metrics/index.md
```diff
@@ -10,37 +10,61 @@ enterprise: false
...
+For more information about the list of metrics that are automatically collected by DC/OS, read [Metrics Reference](/1.13/metrics/reference/) documentation.
 
-## Upgrading from 1.11
-DC/OS 1.12 includes an updated `statsd` server implementation for application metrics. The `statsd` update fixes an issue with the `statsd` server implementation in 1.11, which treated all application metrics as gauges, regardless of `statsd` type. 
+# Upgrading from 1.12
+DC/OS 1.13 includes an updated `StatsD` server implementation for application metrics. The `StatsD` update fixes an issue with the `StatsD` server implementation in 1.12, which treated all application metrics as gauges, regardless of `StatsD` type. 
 
-Dashboards and alerts that rely on counters, histograms, or sets behave differently in 1.12 than in 1.11 as follows:
-- Gauges report the last received value. There is no change from 1.11 functionality. 
+Dashboards and alerts that rely on counters, histograms, or sets behave differently in 1.13 than in 1.12 as follows:
```
gracedo: Shouldn't this be `...differently in 1.13 than in 1.11` - 1.12 and 1.13 are the same, the change is from 1.11

pages/1.13/metrics/index.md
```diff
@@ -10,37 +10,61 @@ enterprise: false
...
+- Gauges report the last received value. There is no change from 1.12 functionality.
```
gracedo: `1.11` perhaps?

pages/1.13/metrics/index.md
philipnrmn: Nice spot. This looks like a victim of a search-and-replace. This section is only relevant to 1.12, and should not be present in the 1.13 docs.

pages/1.13/metrics/index.md
```diff
@@ -10,37 +10,61 @@ enterprise: false
...
+## Using Telegraf
+Telegraf collects application and custom metrics through the `dcos_statsd` plugin. A dedicated `StatsD` server is started for each new task. Any metrics received by the `StatsD` server are tagged with the task name and its service name. The address of the server is provided by environment variables (`STATSD_UDP_HOST` and `STATSD_UDP_PORT`). 
+
+<p class="message--note"><strong>NOTE: </strong>When a task is completed, metrics that are emitted but not gathered by Telegraf will be discarded.</p>
+
+The metrics collected by `dcos_statsd` are gathered every 30 seconds. The task must run for at least 30 seconds to ensure a task's metrics are gathered.
+
+For more information about the list of metrics that are automatically collected by DC/OS, read [Metrics Reference](/1.13/metrics/reference/) documentation.
 
-## Upgrading from 1.11
```
philipnrmn: Let's remove this entire section - it is relevant to users upgrading 1.11 -> 1.12.

It is not true for users upgrading 1.12 -> 1.13

PR #2265, opened 2019-05-20
Simplify the package registry documentation

takirala:
## Jira Ticket
[DCOS-50426](https://jira.mesosphere.com/browse/DCOS-50426) Updates for 1.13: Structure and Commands. Also add documentation on parts that were meant for internal usage only earlier.

* After being in multiple customer calls and discussions with SA's on how to use the documentation, I think we can do some re-structuring of documentation.
* Parts of the documentation (such as how to build a `.dcos` was never made public - although the tools were public. This was fine as it was supposed to be mostly for internal usage). This PR fixes that. 
* There are many parts of this documentation that can skipped as per customer requirements. In fact the entire installation of package registry can be simplified to be a one liner with `dcos registry activate --override`. By organising the documentation in to multiple sections, it was confusing the users as to see what is optional vs what is mandatory.

## Checklist
- [ ] Make sure you change all affected versions (e.g. 1.10, 1.11, 1.12, 1.13). Only supported 1.12 and above. We can update 1.12 after we are happy with the structure suggested in 1.13
- [ ] Test all commands and procedures where applicable.


takirala:
@jpe442 happy to discuss this offline if needed. Also, please add more reviewers who would like to have a say in this.

jpe442:
@takirala - Do you think we need both levels 'Managing Package Registry' and then 'Package Registry' as a leaf below, or is the 'Package Registry ' leaf redundant?

takirala:
@jpe442 yes the package registry leaf is redundant. we can just rename the top level leaf to `Package Registry`.

PR #2316, opened 2019-06-07
[DS] Remove older, unsupported frameworks out of sight

samvantran:
Our official policy is to support N-1 major versions of any given data service framework. This PR marks older versions invisible so they do not appear in our [service docs website](https://docs.mesosphere.com/services). 

At the end of the day, there should only be 2 versions displayed.

### DIFF
--- a/pages/services/confluent-kafka/2.3.0-4.0.0e/index.md
+++ b/pages/services/confluent-kafka/2.3.0-4.0.0e/index.md
@@ -3,7 +3,7 @@ layout: layout.pug
 navigationTitle: Confluent Kafka 2.3.0-4.0.0e
 excerpt:
 title: Confluent Kafka 2.3.0-4.0.0e
-menuWeight: 10
+menuWeight: -1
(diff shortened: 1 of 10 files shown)


ck4adventure:
Hey @samvantran and @zmalik, it looks like we are ready to deprecate these versions?? If so, I will actually delete the folders listed rather than just hiding them via menuweight. Like all things git, they will still be accessible for recovery later.

SarahStegall:
Do we have a JIRA ticket in support of this PR? The policy of the docs team is to track work through JIRA. Do you want me to make a ticket?

SarahStegall:
Most of these changes were already made in DCOS-51550.

SarahStegall:
The only change in this PR which was not made as part of DCOS-51550 was the removal of Spark 2.7.0. Rather than duplicate this effort, I am going to close this PR and move all subsequent work to DCOS-51550, as it is the policy of the docs team to work from JIRA tickets.

PR #2357, opened 2019-06-21
Revert "CK Update to dynamic links"

ck4adventure:
Reverts mesosphere/dcos-docs-site#2356


ck4adventure:
Reverting in order, first dynamic, then main PR

PR #2432, opened 2019-07-24
[DCOS-55313] Cassandra s3 compatible Backup Docs update.

viivek46:
## Jira Ticket

Resolves  [DCOS-55313](https://jira.mesosphere.com/browse/DCOS-55313)

## Checklist
- [x] Make sure you change all affected versions (e.g. 1.10, 1.11, 1.12, 1.13).
- [x] Test all commands and procedures where applicable.
- [x] Add [redirects](https://github.com/mesosphere/dcos-docs-site/wiki/Redirects) if you are moving a page.


pages/services/cassandra/2.5.0-3.11.3/operations/index.md
```diff
@@ -229,8 +229,8 @@ Restoring cluster data is similar to backing it up. The `restore-s3` plan assume
 
 Optional parameters:
 - `S3_ENDPOINT_URL`: URL of S3 compatible storage. If you want to restore from S3 compatible storage, please provide this parameter in `restore-s3` plan.
-- `AWS_SESSION_ID`: It may also be necessary to set the `AWS_SESSION_ID` depending on how you authenticate with AWS.
-- `AWS_SESSION_TOKEN`: It may also be necessary to set the `AWS_SESSION_TOKEN` depending on how you authenticate with AWS.
```
samvantran: why remove the code format? IMO it looked better before

viivek46: well over site it was not looking good. 
https://docs.mesosphere.com/services/cassandra/2.6.0-3.11.4/operations/#backing-up-to-s3
so I changed it.
but if you want me to revert it back i will do.

viivek46:
> I hope you don't mind I committed right to your branch @viivek46 to fix the service docs that appear on the site.
> 
> Right now we're only showing docs for `2.6.0-3.11.4` and `2.5.0-3.11.3` but we should be showing `2.4.0-3.0.16` as that is n-1 major version.
> 

ofcourse not, please updated it as you need.

pages/services/cassandra/2.5.0-3.11.3/operations/index.md
samvantran: I see what you're getting at. Looks confusing to use the actual value twice. How about this as a compromise? 
```
- `AWS_SESSION_ID`: It may also be necessary to set the session ID depending on how you authenticate with AWS.
- `AWS_SESSION_TOKEN`: It may also be necessary to set the session token depending on how you authenticate with AWS.
```

viivek46: :+1: yea that will work, thanks @samvantran

ck4adventure:
Ok, so just to confirm @samvantran and @viivek46 that you want all three versions of cassandra showing?? 2.5.0-3.11.3 having a menuweight of -1 hides it from the nav menu but still builds the directory for direct access. All pages should have positive number menuweights. If they are not meant to be shown they should be deleted.

samvantran:
In that case let's delete `2.5.0-3.11.3`. We only want to display `3.0.16` and `3.11.4`

viivek46:
@ck4adventure i think this PR is not going to work as whole structure of staging is changed. 
could you please confirm me that below is new path for cassandra docs: 
dcos-docs-site/pages/mesosphere/dcos/services/cassandra/

samvantran:
yeah @viivek46 the new directory structure matches the URL of new docs pages: https://docs.d2iq.com/mesosphere/dcos/services/cassandra/

so please move pages under `/mesosphere/dcos/services/cassandra`

viivek46:
new PR  #2496

PR #2592, opened 2019-10-21
Release notes for 2.0.

SarahStegall:
## Jira Ticket
https://jira.mesosphere.com/browse/DCOS-60191

Add changes from Changelog, JIRA tickets, etc. to Release Notes for 2.0.

### DIFF
--- a/pages/mesosphere/dcos/2.0/cli/command-reference/dcos-node/dcos-node-diagnostics-create/index.md
+++ b/pages/mesosphere/dcos/2.0/cli/command-reference/dcos-node/dcos-node-diagnostics-create/index.md
@@ -12,7 +12,7 @@ model: /mesosphere/dcos/2.0/data.yml
 # Description
 The `dcos node diagnostics create` command allows you to create a diagnostics bundle.
 
-**This command is deprecated since DC/OS 2.0; please use `dcos diagnostics create` instead.**
+<p class="message--warning"><strong>WARNING: </strong>This command is deprecated since DC/OS 2.0; please use <a href="/mesosphere/dcos/2.0/cli/command-reference/dcos-diagnostics/dcos-diagnostics-create/"><tt>dcos diagnostics create</tt></a> instead.</p>
 
 # Usage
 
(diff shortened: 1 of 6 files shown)


SarahStegall:
> (1) not sure if the link it working (it doesn't look right here): [https://downloads.dcos.io/dcos/stable/2.0.0/dcos_generate_config.sh"]Download](https://downloads.dcos.io/dcos/stable/2.0.0/dcos_generate_config.sh%22%5DDownload) <-- I wonder if it needs a space between "]" and "Download"

The link does not work because the URL is not yet active. Typically the URL for the download page is not live until the release date; it is usually the very last change made to the Release Notes and is done on the GA date.

> (2) Under "Release Summary" we may want to capitalize the "I" in "Known Issues and Limitations"

Done. Thanks.

> (3) We don't always use a period after the Jira key. For example, (DCOS-54186). vs. (DCOS-53654)

Thanks. Fixed.

> (4) Sometimes we explain things as "DC/OS has a new ..." or "DC/OS supports..." and other times we use past-tense verbs such as "Added" "Created" "Updated". Maybe it doesn't matter but it's inconsistent.

Noted and fixed. Thanks.

SarahStegall:
> under issues fixed in this release, third bullet. dc/os has a new container debug endpoint, and [the] diagnostic bundle includes the debug endpoint tracking data for stuck task[s]. (or for [a] stuck task.)

Thank you. Fixed.


> 
> also, we talked about whether release notes should be in present or past tense (at least i did with casie). i had heard that it should be in present tense. either way, i believe it should be in the same tense throughout, and there are places where there is present tense, and places where there is past tense.

Wherever practical, I have made everything present tense. The exception is in the Release Summary, where the description of updates to core components like Marathon and Mesos have always been described in past tense. I continued that practice here.

SarahStegall:
@cmastr "Previous releases only point to the last major releases?"
The list of previous releases should point to the last point release, so that the URL for the last release of 1.13 would point to the release notes for 1.13.5.

SarahStegall:
@cmastr Comments addressed.

PR #2682, opened 2019-12-17
correct url in spark submit example for DSE

Gupta-Amrit:
## Jira Ticket
https://jira.mesosphere.com/browse/COPS-5700


cmastr:
Going to wait on response from Support before we merge this.

cmastr:
deleting this change as we will be fixing the location of the download to match the existing URL.

PR #2754, opened 2020-02-11
[COPS-5739] - add a bugfix to the release notes of 1.13.6

pierrebeitz:
https://jira.d2iq.com/browse/COPS-5739 is requesting to let customers know about a bugfix in DCOS-UI
that let's users correctly handle the deletion and re-addition of
secrets.


cmastr:
Thank you @pierrebeitz! Pushing to production.

PR #2770, opened 2020-02-19
[kommander] Delete istio 'coming soon' page

samvantran:
Empty page in the tutorial section for a feature 'Coming Soon' is misleading. I recommend deleting it and adding an actual tutorial when we have something worked out.

Note: I'm creating a PR off @ck4adventure's branch that includes a number of kommander doc fixes.


cmastr:
I agree. Need to check with Jackson and Chris though because this was a pretty specific ask.

jaxesn:
This is ok to remove. We've removed it from the kommander UI as well since
we just don't really know when it will come.

PR #2800, opened 2020-03-10
Add instructions for using 3rd party Docker volume driver.

qianzhangxa:
## Jira Ticket
[D2IQ-26077](https://jira.d2iq.com/browse/D2IQ-26077)

This PR adds instructions for using 3rd party Docker volume driver.

### DIFF
--- a/pages/mesosphere/dcos/1.13/storage/external-storage/index.md
+++ b/pages/mesosphere/dcos/1.13/storage/external-storage/index.md
@@ -128,6 +128,24 @@ Apps that use external volumes can only be scaled to a single instance because a
 
 If you scale your app down to 0 instances, the volume is detached from the agent where it was mounted, but it is not deleted. If you scale your app up again, the data that was associated with it is still be available.
 
+# Using 3rd party Docker volume driver
+
+If you want to use a 3rd party Docker volume driver rather than REX-Ray (e.g., [NetApp Trident](https://github.com/NetApp/trident)), you will need to take the following steps on each agent node (Trident is used as an example in the steps below):
+
+1. Install the volume driver as a Docker plugin.
+    ```
+    $ docker plugin install netapp/trident-plugin:19.10 --alias netapp --grant-all-permissions
+    ```
+
+1. Register the volume driver to Docker plugin directory. Refer to [Docker Plugin API](https://docs.docker.com/engine/extend/plugin_api/#plugin-discovery) to learn how Docker discovers plugins.
+
+    ```
+    $ sudo mkdir -p /etc/docker/plugins/
+    $ sudo bash -c 'echo "unix:///run/docker/plugins/<plugin-id>/netapp.sock" > /etc/docker/plugins/netapp.spec'
+    ```
+
+Now the volume driver is ready for you to use it in DC/OS. When you create an application, you need to set the option `external.options["dvdi/driver"]` to the name of the volume driver (e.g., `netapp`).
+
 # Potential issues
 
 *   You can assign only one task per volume. Your storage provider might have other limitations.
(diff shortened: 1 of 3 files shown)


corbinpacheco:
@qianzhangxa Rancher is one of our main competitors.  I was wondering if we could use a different docker driver as an example.  Netapp is a commonly requested driver.  Is it possible for you to test integrating that one with your procedure and using that as the example?  https://github.com/NetApp/trident

qianzhangxa:
@corbinpacheco Thanks for the comments! I have updated the PR to use NetApp as example.

cmastr:
will wait for Corbin to approve changes then I'll merge

vjone:
So I was trying to execute the instructions, when it says "Now the volume driver is ready for you to use it in DC/OS. When you create an application, you need to set the option external.options["dvdi/driver"] to the name of the volume driver (e.g., netapp)." I think we need a screenshot of where/how to set that option explicitly. For me, it's unclear.

vjone:
also, I executed:

'docker plugin install netapp/trident-plugin:19.10 --alias netapp --grant-all-permissions'

that was on a private agent dc/os node and got the below output error:

[root@ip-172-16-13-232 ~]# docker plugin install netapp/trident-plugin:19.10 --alias netapp --grant-all-permissions
19.10: Pulling from netapp/trident-plugin
9ae4298deace: Download complete
Digest: sha256:494846e67f494e0c886920b58095d01c96062148e4d7b3606bc221c6dade3df9
Status: Downloaded newer image for netapp/trident-plugin:19.10
Error response from daemon: OCI runtime create failed: container_linux.go:345: starting container process caused "process_linux.go:430: container init caused \"rootfs_linux.go:58: mounting \\\"/etc/netappdvp\\\" to rootfs \\\"/var/lib/docker/plugins/2382440af3ff3c5e4779c252d5425404a80d045a629bb5848b7c3d11089ba1e0/rootfs\\\" at \\\"/etc/netappdvp\\\" caused \\\"stat /etc/netappdvp: no such file or directory\\\"\"": unknown

I'm assuming this is incorrect? What should the correct output look like?

qianzhangxa:
> So I was trying to execute the instructions, when it says "Now the volume driver is ready for you to use it in DC/OS. When you create an application, you need to set the option external.options["dvdi/driver"] to the name of the volume driver (e.g., netapp)." I think we need a screenshot of where/how to set that option explicitly. For me, it's unclear.

Actually this is the second time that the option `external.options["dvdi/driver"]` is mentioned in this doc, the first time is here: https://docs.d2iq.com/mesosphere/dcos/2.0/storage/external-storage/#volume-configuration-options. So I think if users read this doc from the beginning (rather than just read the new section that I introduced in this PR), then they should know what the option `external.options["dvdi/driver"]` means and where to set it (there is already a sample application JSON in this doc).

qianzhangxa:
> 'docker plugin install netapp/trident-plugin:19.10 --alias netapp --grant-all-permissions'

This is just an example, before you execute it, you need to have the supported NetApp storage backend and have a configuration file to specify how to connect to the storage backend, see https://github.com/NetApp/trident/blob/master/docs/docker/deploying.rst for more details. I just want to put the general steps in our doc (like `docker plugin install ...`) but not the plugin specific steps.

vjone:
Fair point that each 3rd party storage driver is going to have different instructions. I would just say that, we should probably explicitly say that in big bold letters, because even the example with NetApp Trident here is incomplete without going to https://github.com/NetApp/trident/blob/master/docs/docker/deploying.rst and creating the config.json file from those instructions. 

And as you said, you actually need the true netapp back-end running before just doing this example, so explicitly saying that is required to do this example successfully would be clarifying vs just trying to do the example verbatim and failing without a real netapp backend.

PR #2824, opened 2020-03-31
UI Updates

ck4adventure:
It needed a visual pass. Before is on the left, after is on the right.

Top level pages - Increased spacing between jumbotron and content, locked jumbotron font size at xxl
Increased spacing between jumbotron and services page content, set to left aligned and all caps on categories to help distinguish more clearly

Centered version picker on content pages

one

two

three


ck4adventure:
live on docs-christina.d2iq.com for preview

cmastr:
looks better!

PR #2828, opened 2020-04-01
Publish DCOS 2.1 

ck4adventure:
DC/OS 2.1 is currently blocked from building on staging/master. This allows local preview work. Once the 2.1 content is in and approved, simply merge this PR into staging and promote. The block is released by deleting the lines. The build will pick up the change and add 2.1 in.

### DIFF
--- a/config.json
+++ b/config.json
@@ -2,15 +2,13 @@
   "master": {
     "DO_NOT_BUILD": [
       "test/**",
-      "mesosphere/dcos/1.14/**",
-      "mesosphere/dcos/2.1/**"
+      "mesosphere/dcos/1.14/**"
     ],
     "DO_NOT_INDEX": []
   },
   "staging": {
     "DO_NOT_BUILD": [
-      "test/**",
-      "mesosphere/dcos/2.1/**"
+      "test/**"
     ],
     "DO_NOT_INDEX": [
       "^cn"
@@ -46,7 +44,7 @@
       "mesosphere/dcos/1.11/**",
       "mesosphere/dcos/1.12/**",
       "mesosphere/dcos/1.13/**",
-      "mesosphere/dcos/2,1/**"
+      "mesosphere/dcos/2.1/**"
     ],
     "DO_NOT_INDEX": []
   },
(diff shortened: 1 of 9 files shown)


pages/mesosphere/dcos/2.1/release-notes/2.1.0/index.md
```diff
@@ -2,13 +2,14 @@
...
-menuWeight: 5
+beta: true
+menuWeight: 0
 render: mustache
 beta: true
 model: /mesosphere/dcos/2.1/data.yml
 excerpt: Release notes for DC/OS 2.1.0, including Open Source attribution, and version policy.
 ---
-DC/OS 2.1.0 was released on Day Month, 2020.
+DC/OS 2.1.0 Beta was released on 27 April, 2020.
```
vinodkone, line 12: DC/OS 2.1.0 Beta4 was released on 27 April, 2020.

pages/mesosphere/dcos/2.1/release-notes/2.1.0/index.md
```diff
@@ -24,8 +25,8 @@ This release provides new features and enhancements to improve the user experien
 If you have DC/OS deployed in a production environment, see [Known Issues and Limitations](#known-issues-and-limitations) to see if any potential operational changes for specific scenarios apply to your environment.
 
 - Updated DC/OS UI to master+v2.150.2.
-- Updated to Mesos ??
-- Updated Marathon to 
+- Updated to Mesos ???
```
vinodkone, line 28: This page seems out of data. Lets use data from https://github.com/dcos/dcos/blob/master/CHANGES.md as the starting point for release notes?

config.json
```diff
@@ -46,7 +44,7 @@
       "mesosphere/dcos/1.11/**",
       "mesosphere/dcos/1.12/**",
       "mesosphere/dcos/1.13/**",
-      "mesosphere/dcos/2,1/**"
+      "mesosphere/dcos/2.1/**"
```
vinodkone, line 47: why is 2.0 not specified here or anywhere else in this page?

pierrebeitz: in my understanding the purpose of listing versions under `DO_NOT_BUILD` is to speed up builds. `2.0` likely could be added in all of those branches. on the other hand some of the branches (and respective pipelines) and the reference to `1.14` are known to be obsolete, so this file could need some attention in general. 

also "mesosphere/**" might be an easier glob for ksphere's configs.

cmastr: This is right Pierre. We need to add 2.0 only to the other product lists. But that's not the intent of this PR.

PR #2827, opened 2020-03-31
Konvoy Versioning Code Changes

ck4adventure:
This PR is only meant to test all the code changes for UX and display. 

DO NOT EDIT KONVOY CONTENT - IT"S OUTDATED ON PURPOSE

the 1.5 docs are also completely blocked from building on prod sites

PR #2844, opened 2020-04-16
konvoy multiple versions

dkoshkin:
## Jira Ticket
https://jira.d2iq.com/browse/D2IQ-66900

Follow up to https://github.com/mesosphere/dcos-docs-site/pull/2827.

It will be easiest to review by going through each commit.

* Copied the existing docs from `latest/` to ~`1.5/`~ `1.5-beta/`.
* Moved all the docs from Konvoy `release-1.4` branch to `1.4/`
* Renamed the titles to make it clearer what versions are in the dropdown
* Removed the beta tags and set it on the main index page for ~`1.5/`~ `1.5-beta/`, that whole version is in beta.
* Added a redirect from `/ksphere/konvoy/latest/` to `/ksphere/konvoy/1.5-beta/`
* Renamed `1.5` to `1.5-beta`.

You can see the docs at http://cf5146aa.ngrok.io/ksphere/konvoy/  (will be available next 8 hours or when I close the lid)


docker/nginx/redirects-307.map
```diff
@@ -1,4 +1,5 @@
 ~^/mesosphere/dcos/latest/(.*)$ /mesosphere/dcos/2.0/$1;
+~^/ksphere/konvoy/latest/(.*)$ /ksphere/konvoy/1.5/$1;
```
samvantran, line 2: It looks like we might need more than just this for the redirects to work properly. I tried to substitute URL with `latest` and the left navbar loses the versions and it seems to point back to 1.4 actually, see screenshot 

Screen Shot 2020-04-16 at 3 12 27 PM

I saw this both on your hosted site and my local build

samvantran: Another thing is the left navbar will show `-beta` in the dropdown but not the main top left display so when looking through different pages you won't immediately recognize that these are beta docs 

Screen Shot 2020-04-16 at 3 17 13 PM

dkoshkin:
https://github.com/mesosphere/dcos-docs-site/pull/2844#discussion_r409880371

@samvantran I am not using docker, thats why its not being redirected on mine?
Did you run it in a container?

dkoshkin:
https://github.com/mesosphere/dcos-docs-site/pull/2844#discussion_r409882240

@samvantran that "Konvoy 1.5" is driven by the directory name.
One option is we could create `1.5-beta` directory instead of `1.5`, in that case we would have to check if $VERSION (konvoy tag) contains "beta" and then include the `-beta` suffix if it does.

dkoshkin:
image

Added the `-beta` suffix which is reflected in the dropdown and the URL, that makes more sense.

The script will just need to become more sophisticated to properly handle the tag.

pages/ksphere/konvoy/1.4/quick-start/index.md
```diff
@@ -0,0 +1,235 @@
...
+1.  Check the Kubernetes client version. Many important Kubernetes functions **do not work** if your client is outdated. You can verify that the version of `kubectl` you have installed is supported by running the following command:
+
+    ```bash
+    kubectl version --short=true`
+    ```
+
+1.  Download and extract the Konvoy package. You will need to download and extract the Konvoy package tarball. There are a couple of ways to get the tarball:
+
+    - You can request a free trial from D2iQ [here](https://d2iq.com/solutions/ksphere/konvoy#request-free-trial). We will send you a link for the tarball.
+    - If you are a D2iQ customer, you can ask your account representative for a trial license.
```
jenafermiller, line 59: If they are an existing Konvoy customer they can download it from the support portal (support.d2iq.com)

jenafermiller, line 59: If they are an existing Konvoy customer they can download it from the support portal (support.d2iq.com)

dkoshkin: @russurquhart1 can you please create a ticket to fix this.
This PR is only for the structure, not the content.

cmastr: Yeah, this should match the other pages where we talk about this. Or possibly, we should get rid of these references and just point to the master topic. I'd say don't hold this PR for that though, let's open a JIRA>

PR #2859, opened 2020-04-21
Ignore more DC/OS documentation for the dispatch branch build

jbarrick-mesosphere:

Optimize the `dispatch` branch build so that our dispatch CI builds do not have to build all of the DC/OS documentation.


cmastr:
@jbarrick-mesosphere if you ignore these, then they won't build at all. The docs site needs to build all the docs every time, otherwise we won't have DCOS docs on the website anymore.

PR #2872, opened 2020-04-28
chore: fix the revision-plugin

pierrebeitz:
# ⚠️ EDIT

1. the numbers and depicted speeds you can see in the screencasts below changed because #2873 merged.
2. you'll only want to review the last commit (`chore: fix the revision-plugin`) here. the other commits came here due to rebasing and are already approved/merged.

----------------------

to speed up builds, we now make use of the revision-plugin and
assume that all files that need processing are changed after the
markdown-step. the revision-plugin will then generate a hash out of the
current state of that file and not apply further transformers in case
the hash is the same as in the previous run.

note that the official revision plugin would introduce caching issues as it does not care about changes in frontmatter.

here are two (almost) identical cycles of starting the server, making a change and waiting for that change to be visible in the browser: 

## Before (96 secs total; time until a change can be seen in browser: 34 secs)

before_small

## After (44 secs total, time until a change can be seen in browser: 11 secs)

after_small

## Checklist
- [x] Make sure you change all affected versions (e.g. 1.10, 1.11, 1.12, 1.13).
- [x] Test all commands and procedures where applicable.
- [x] Add [redirects](https://github.com/mesosphere/dcos-docs-site/wiki/Redirects) if you are moving a page.

### DIFF
--- a/index.js
+++ b/index.js
@@ -334,6 +334,7 @@ if (process.env.NODE_ENV === 'development' && RENDER_PATH_PATTERN) {
             [`pages/${RENDER_PATH_PATTERN}/*`]: '**/*.{md,tmpl}',
             'layouts/**/*': '**/*.pug',
         },
+        livereload: true,
     }));
     CB.use(timer('CB: Watch'));
 }
--- a/index.js
+++ b/index.js
@@ -1,5 +1,6 @@
 // Packages
 const fs = require('fs');
+const minimatch = require('minimatch');
 const Metalsmith = require('metalsmith');
 const markdown = require('metalsmith-markdownit');
 const layouts = require('metalsmith-layouts');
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -215,6 +209,12 @@ CB.use(shortcodes({
 }));
 CB.use(timer('CB: Shortcodes'));
 
+// Don't rebuild files that have not been touched
+if (process.env.NODE_ENV === 'development') {
+  CB.use(revision);
```
pierrebeitz, line 214: this looks at the file in its current state in the pipeline and removes it from further processing in case it looks identically (reasoning: "nothing changed, so we don't need to run all the processors on every file every time")

i arbitrarily placed that after shortcodes, so changes to shortcode-replacements do invalidate the cache (also the next processor - markdown - is the first one to be quite slow).

plugins/metalsmith-revision.js
```diff
@@ -1,85 +1,53 @@
...
+const fs = require("fs");
+const crypto = require("crypto");
+
+const checksum = (str) =>
+  crypto.createHash("md5").update(str, "utf8").digest("hex");
+
+// attrubutes that are not used to generate a cache-hash.
+const attrsToIgnore = ["stats", "model", "filename"];
+
+module.exports = function (files, metalsmith, done) {
```
pierrebeitz, line 11: ### TL;DR

* if any `meta` a.k.a "frontmatter" changed, we rebuild everything
* if some `contents` changed, we rebuild only those pages

----------------

longer story: we place a `.revision` file in the root folder (as that is what the previous version did). that file contains two dicts - `contents` and `meta`. `contents` maps each file to a checksum of its contents, `meta` maps each file to a checksum of its frontmatter-entries. this is necessary to e.g. invalidate this cache in case you change the `title` or `navigationTitle` or whatnot, as such change needs to propagate to every page (cause of the way out layouts work).

as some `meta` entries are not something we specify, we strip them via `attrsToIgnore`.

-------------

groking this is kinda hard i guess, so feel free to reach out via zoom and i walk you through, if you really want to understand what's going on here.

cmastr:
looks good! Changed to staging branch. Would like a couple more looks before we merge.

PR #2881, opened 2020-04-28
[DCOS_OSS-5953] Removing Install as everything is in the quick start …

msdolbey:
…guide

This is no longer a workflow that users would use.
When this was written, it wasn't clear how kommander would be distributed
One possibility at the time, is if you download konvoy maybe you don't get it?
But, the fact is we bundle a limited version of kommander with konvoy that you can use without a license for up to 2 clusters
Well, one cluster plus konvoy
So, it turns out almost no one would install kommander this way

Relevant slack conversation about this topic: 
https://mesosphere.slack.com/archives/CQL8757S7/p1588105340311000

## Checklist
- [ ] Make sure you change all affected versions (e.g. 1.10, 1.11, 1.12, 1.13).
- [ ] Test all commands and procedures where applicable.

@catsouthard  @cmastr @russurquhart1 @brandonc  - could you give a review?

I ran this on the docs site locally, and I'm pretty sure I'm not missing any links or anything.


juliangieseke:
rebased this on top of new beta docs, its removed in both folders now.

cmastr:
See slack thread, but here's my response:

Ok, coming in a bit late on this. We should not only have a quick start from a customer perspective. They are not going to be looking for "quick start" when they google "Konvoy Install" or whatever. SO I understand the intention here. But if we're gonna delete the Install topic, we then need to rename the Quickstart to "install".
Plus, I hate the term Quickstart in general. Quick for whom exactly? Is quick 1 minute or 1 hour? Let's just not use this. @Russ Urquhart

Casie Oxford  < 1 minute ago
So will need to get some updates to this PR before I'm comfortable merging this one.

msdolbey:
@cmastr 
closing this one in preference to this one: 
https://github.com/mesosphere/dcos-docs-site/pull/2888
- going to continue the conversation there.

PR #2897, opened 2020-05-04
ksphere: bring back redirects from latest/

dkoshkin:
These were incorrectly removed in https://github.com/mesosphere/dcos-docs-site/pull/2828/files#diff-ffadabb9049686535404014ce18e6a3bL1-L4 and broke the `/latest` redirects.

Bringing it back.


cmastr:
Sorry, I was checking those very carefully with the rebase, but apparently still didn't get it right. I can merge this now.

PR #2903, opened 2020-05-05
Remove Portworx from Partner Solutions.

russurquhart1:
## Jira Ticket

https://jira.d2iq.com/browse/D2IQ-67468


samvantran:
Looks good but this should be pointing to head branch `staging`, not `master`

dkoshkin:
Please also add a description on why this change is being made, for future reference.

PR #2907, opened 2020-05-07
Add Kubeflow-scaffold

pierrebeitz:
# ⚠️ do not merge (yet)!

### depends on https://github.com/mesosphere/dcos-docs-site/pull/2908.

review and merge https://github.com/mesosphere/dcos-docs-site/pull/2908 first. that PR is independent of kubeflow-docs. 

-------------

# Screenshot

Just a first shot on how we might want to add a new tile. I could also imagine a 3/2-layout, where 3 tiles are on the upper row and two tiles move to a second row or a layout where we present the products vertically. 

D2iQ Docs 2020-05-07 18-19-59

EDIT: there now also is a semi-functional kubeflow-section. by semi-functional i mean that a lot of layouts seem to be overly complicated and thus i'm not sure whether headers work the the way they should.

### DIFF
--- a/entry.js
+++ b/entry.js
@@ -23,8 +23,3 @@ require('./js/msphere-menu.js');
 require('./js/ksphere-menu.js');
 require('./js/localization.js');
 require('./js/services-landing.js');
-if ( ENV === "development" ){
-  var script = document.createElement('script');
-  script.src = 'http://localhost:35729/livereload.js';
-  document.getElementsByTagName('head')[0].appendChild(script);
-}
--- a/index.js
+++ b/index.js
@@ -115,8 +115,9 @@ MS.metadata({
     dcosDocsLatest: '2.0',
     dcosCNDocsLatest: '2.0',
     konvoyDocsLatest: '1.4',
-    kommanderDocsLatest: '1.0',
     dispatchDocsLatest: '1.1',
+    kommanderDocsLatest: '1.0',
+    kubeflowDocsLatest: '1.0',
 });
 
 // Source
@@ -151,6 +152,8 @@ const neededToBuildMainMenu = [
   `ksphere/konvoy/${MS._metadata.konvoyDocsLatest}/index.md`,
   "ksphere/kommander/index.md",
   `ksphere/kommander/${MS._metadata.kommanderDocsLatest}/index.md`,
+  "ksphere/kubeflow/index.md",
+  `ksphere/kubeflow/${MS._metadata.kubeflowDocsLatest}/index.md`,
 ]
 if (process.env.NODE_ENV === 'development' && RENDER_PATH_PATTERN) {
   MS.use((files, _, done) => {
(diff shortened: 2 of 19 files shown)


danielone:
@pierrebeitz do you feel we can fit a 5th tile here without scroll on a laptop size screen? how much would the current tile size need to be reduced?

would it be possible to prototype putting the tile graphic to the left of each text block below please? this is an alt layout @cmastr and I are discussing ?

pierrebeitz:
@danielone if you mean the horizontal scrollbar: nice catch! unfortunately it's not an issue i introduced in this PR but we also have that live. here's a fix: https://github.com/mesosphere/dcos-docs-site/pull/2914.

so the tile sizes will stay the same as in the current state of this PR. they vary between 768px and 1024px.

### min 

D2iQ Docs 2020-05-10 05-31-42


### max

D2iQ Docs 2020-05-10 05-30-12


# alternative layout prototype

i tried to do something sensible with my limited layouting-skills. i feel that this does not look right yet, but think it might be a valid option with some more of your input:

D2iQ Docs 2020-05-10 06-09-30

here's the commit in case you want to check it out and tweak stuff yourself in the `scss`-file: https://github.com/mesosphere/dcos-docs-site/commit/daada0962e55e5d6e807e6dea68ee05cfa845170

cmastr:
> @danielone if you mean the horizontal scrollbar: nice catch! unfortunately it's not an issue i introduced in this PR but we also have that live. here's a fix: #2914.
> 
> so the tile sizes will stay the same as in the current state of this PR. they vary between a screen size of 768px and 1024px as follows:
> 
> ### min
> D2iQ Docs 2020-05-10 05-31-42
> 
> ### max
> D2iQ Docs 2020-05-10 05-30-12
> 
> # alternative layout prototype
> i tried to do something sensible with my limited layouting-skills. images are vertically centered and are hidden below laptop-size. i feel that this does not look right yet, but think it might be a valid option with some more of your input:
> 
> D2iQ Docs 2020-05-10 06-09-30
> 
> here's the commit in case you want to check it out and tweak stuff yourself in the `scss`-file: [daada09](https://github.com/mesosphere/dcos-docs-site/commit/daada0962e55e5d6e807e6dea68ee05cfa845170)

This looks pretty good! And it would save us from the space issues we may have in the future if we have additional products.

danielone:
@pierrebeitz  thanks for trying that out!! It looks like @cmastr likes the general direction too. 
We may need to tweak the image sizing and ensure same alignment for image for each section, as well keep description text length more balanced between each but looks like we have a good path if Anne Marie also on board. 

In the tile region above, ideally we show some sort of filter to filter down the choices by sphere inspired by d2iq.com site:
https://share.getcloudapp.com/E0ubLNB1

@cmastr good to chat more about with AnneM at our meeting coming up.

pierrebeitz:
let me know if something like a live-tweaking-session would be helpful. 

i'm on PTO next week though.

cmastr:
Ok, so I found out yesterday that we are going to need to add Conductor soon as well. So we definitely need to go the vertical route I think. Let's not add Conductor in this PR though.

cmastr:
Has lots of merge conflicts now and we've decided to go in another direction, so suggestion is to close this in favor of upcoming design work.

PR #2908, opened 2020-05-07
Livereload for layouts and assets

pierrebeitz:
1. the livereload-script is moved to partials/scripts as that seems to
be included in more pages (like the landing page)
2. the pug- and consolidate-cache are cleared inbetween runs. we want to
keep the cache within a run though as it massively speeds up things.
3. add hashes for layouts within the revision-plugin to rebuild
everything when a layout changed.
4. `metalsmith-branch` is removed. does not seem to bring any benefit as we ran all files through both pipelines and the processors did the work of filtering our stuff anyway.

# Motivation

you now can edit layouts and see the changes without having to restart the dev environment. layouts and stylesheets/js should livereload now, so we have faster feedback cycles when changing stuff there.

### DIFF
--- a/entry.js
+++ b/entry.js
@@ -23,8 +23,3 @@ require('./js/msphere-menu.js');
 require('./js/ksphere-menu.js');
 require('./js/localization.js');
 require('./js/services-landing.js');
-if ( ENV === "development" ){
-  var script = document.createElement('script');
-  script.src = 'http://localhost:35729/livereload.js';
-  document.getElementsByTagName('head')[0].appendChild(script);
-}
--- a/index.js
+++ b/index.js
@@ -8,7 +8,6 @@ const permalinks = require('metalsmith-permalinks');
 const assets = require('metalsmith-assets');
 const dataLoader = require('metalsmith-data-loader');
 const watch = require('metalsmith-watch');
-const branch = require('metalsmith-branch')
 const serve = require('metalsmith-serve');
 const redirect = require('metalsmith-redirect');
 const webpack = require('metalsmith-webpack2');
@@ -137,11 +136,8 @@ MS.clean(false);
 // Content Branch Pipeline
 //
 
-const CB = branch();
-
 // Start timer
-CB.use(timer('CB: Init'));
-
+MS.use(timer('Init'));
 
 const neededToBuildMainMenu = [
   "index.md",
@@ -157,7 +153,7 @@ const neededToBuildMainMenu = [
   `ksphere/kommander/${MS._metadata.kommanderDocsLatest}/index.md`,
 ]
 if (process.env.NODE_ENV === 'development' && RENDER_PATH_PATTERN) {
-  CB.use((files, _, done) => {
+  MS.use((files, _, done) => {
     Object
       .keys(files)
       .filter(file =>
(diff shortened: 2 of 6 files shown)


entry.js
```diff
@@ -23,8 +23,3 @@ require('./js/msphere-menu.js');
 require('./js/ksphere-menu.js');
 require('./js/localization.js');
 require('./js/services-landing.js');
-if ( ENV === "development" ){
-  var script = document.createElement('script');
-  script.src = 'http://localhost:35729/livereload.js';
-  document.getElementsByTagName('head')[0].appendChild(script);
-}
```
pierrebeitz, line 30: this section has gone to `/layouts/partials/scripts.pug` as the `entry.js` is not included in every page, while the script-partial seems to be.

index.js
```diff
@@ -8,7 +8,6 @@ const permalinks = require('metalsmith-permalinks');
 const assets = require('metalsmith-assets');
 const dataLoader = require('metalsmith-data-loader');
 const watch = require('metalsmith-watch');
-const branch = require('metalsmith-branch')
```
pierrebeitz, line 11: this is usually used to have separate pipeline for stuff. in our case we would actually want something like:

```
const AssetBranch = branch();
const ContentBranch = branch();

AssetBranch.use(filter((file) => isScss(file) || isJs(file))) // <----
AssetBranch.use(assetProcessors)

ContentBranch.use(filter((file) => isMarkdown(file) || isPug(file))) // <----
AssetBranch.use(contentProcessors)

Metalsmith.use(AssetBranch)
Metalsmith.use(ContentBranch)
```

as we don't have those filtering statements, it seems safe to assume that the processors filter our the files they need to process themselves and that we can put everything into a single pipeline. given that the output looks identically (to me) when applying this PR, that seems to hold true.

That's what all the changes including `MS`, `AB`, and `CB` that follow are about. In short: we put the `assets`-processor into the main-pipeline and joined the `watch` expressions.

index.js
```diff
@@ -274,24 +278,31 @@ CB.use(markdown({
...
 // Layouts
-CB.use(layouts({ engine: 'pug', cache: true }));
-CB.use(timer('CB: Layouts'));
+MS.use((...args) => {
+  // we want to use pugs cache for the duration of one (re-)build. we need to clear it inbetween to
+  // have changes to layouts propagate directly. unfortunately there's no convenience for that and we
+  // have to reach into pug and consolidate directly for now.
+  consolidate.clearCache();
+  pug.cache = {};
+  return layouts({ engine: 'pug', cache: true })(...args);
```
pierrebeitz, line 303: we want to keep this `cache: true` as else every layout-step takes ~150ms (tested on `dcos/2.0`). unfortunately that would cache even for subsequent runs and thus ruins live-reload for layouts. so we manually clear the cache for new runs in `pug` and `consolidate` (which wraps pug and brings yet another cache 🤯).

plugins/metalsmith-revision.js
```diff
@@ -51,7 +52,15 @@ module.exports = function (files, metalsmith, done) {
       unchangedFiles.push(file);
     }
   });
+  glob.sync("./layouts/**/*.pug").map((file) => {
+    const layoutHash = checksum(fs.readFileSync(file, "utf-8"));
+    if (oldHashes.layouts[file] !== layoutHash) {
+      rebuildAll = true;
+    }
+    newHashes.layouts[file] = layoutHash;
```
pierrebeitz, line 60: "rebuild everything if a layout changed"

pierrebeitz:
added comments to make this easier to review

PR #2913, opened 2020-05-08
run prettier on js-files and remove linting

pierrebeitz:
today i noticed the instructions in `js/readme.md`: 

> Before pushing any code, lint javascript files with `npm run eslint`

it looks like noone did that in a long time and the current linting rules are not state of the art anymore.

thus i ran `npx prettier --write` on the files in this commit, so they would
be easier to work with. we might want to do this with the markdown-files
at some point in time and also automate this later on.

also eslint has been removed as we already had a whopping 1033 problems
with 876 errors. it didn't seem worthwhile to fix that when there is
something like prettier.js doing this in a matter of seconds.

pbeitz ❐ main ● 2 zsh 2020-05-08 15-23-39

⚠️ this will cause conflicts with other alterations to those files. it currently looks like i'm the only one to suffer from that though :)


cmastr:
I'm going to open a JIRA for the errors. wow a lot!

pierrebeitz:
Oh. Actually this PR removes the linting in favor of using prettier.js and thus the problem of those errors is no more.

pierrebeitz:
thanks @samvantran! that's definitely planned (as hinted in the description). 

i hesitated to make that change directly as i don't know how well things work with node v8 (which is the recommended version to use this repo as per the wiki) and thus wanted to look into that before introducing more fancy tooling. `husky` for example seems to require `node >= 10`. so for now this PR just seeks to make those files easier to work with (as i love prettier handling all whitespace). 

i also plan to reintroduce eslint with a sensible config. unfortunately that needs `node >= 12`.

to conclude: i think this is mergeable in the current state as it provides immediate benefits (for folks who have prettier set up in their editors) without major drawbacks.

cmastr:
@pierrebeitz I merged #2908 now have some conflicting files as you said. Can you take a look?

pierrebeitz:
# ✅

PR #2918, opened 2020-05-12
Fix build when using dispatch or kommander branch

pierrebeitz:
the layouts are written in a way that requires the latest version of any
given doc to be present during build time. the DO_NOT_BUILD sections
exclude dcos 2.0 since 771142bbdce3c03bbf07cdf5f10b84f01d4c7c81.
unfortunately that version is needed to properly build the docs for the
`dispatch` and `kommander` branches.

i also removed the `christina`-branch as nobody seems to use it.

### DIFF
--- a/config.json
+++ b/config.json
@@ -2,14 +2,14 @@
   "master": {
     "DO_NOT_BUILD": [
       "test/**",
-      "mesosphere/dcos/1.14/**", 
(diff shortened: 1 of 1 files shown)


cmastr:
Ah, I understand now! I'll add this to the wiki somewhere. Thank you @pierrebeitz

PR #2958, opened 2020-05-29
Skip indexing kubeflow on master (temporary)

samvantran:
Since kubeflow will be a new product, we need to update algolia and a few places in code to begin indexing. 

This change skips algolia indexing on kubeflow for the time being until we make those changes.


samvantran:
It was all @jaxesn he's the wizard.

PR #3043, opened 2020-06-24
spellchecking Kubeflow

pierrebeitz:
Started to investigate on how we can have a spell check in the docs.
Here's a lightweight method using the markdown-spellcheck package.
https://github.com/lukeapage/node-markdown-spellcheck#cli-options

```
# To generate a report:
npx markdown-spellcheck -a -n -r "pages/ksphere/kubeflow/**/*.md"

# To interactively handle the remaining errors:
npx markdown-spellcheck -a -n -x "pages/ksphere/kubeflow/**/*.md"
```

We're using the following options:
* `-a` Ignore acronyms like NPM. Does not ignore single letters.
* `-n` Ignore numbers. E.g. `0.7` in plain text.
* `-r` Will show a report of all the spelling mistakes that have been found.
* `-x` Suggestions are slow at present, so use this to remove them.

bricks in our way:

- it looks like some files make use of html directly and don't care to use markdown where they actually should
- there are a lot of occasions where things that should be written as  `code` are not. and that's not something that can be changed via the tool itself, so correcting those will be tedious.
- when running on all files (`npx markdown-spellcheck -a -r "pages/**/*.md"`) we currently get ~60k errors.

--------------

To be able to have a fair chance of introducing this, we need to make
this an incremental process. The following methods come to mind:

* Do something with the total number of errors. We could say that it needs to stay in some range. E.g. "a PR should not introduce more than 50 new" or "every PR needs to lower that number" (won't work in real life obviously)
* We can create error free spaces and check that those stay free of errors. `pages/ksphere/kubeflow/**/*.md` might be a good first candidate ;)
* We add all offences to the "global" `.spelling` file (yes, you can create `.spelling` files in subdirectories to only apply for those) and remove entries as we see fit. This might be the fastest method to get rid of actual spelling mistakes as they're the easiest to see in an alphabetically sorted file. Also this would enable a strict CI check right away.
* Your idea...

As an example see the second commit in this PR. It just adds all the offences for "pages/ksphere/kubeflow/**/*.md" - and there seems to be some value in that (`funtions`, `concensus`).

--------------------

BTW: a nifty thing to change all occurrences of something within the kubeflow-folder: 
```
npx fastmod --extensions md "funtions" "functions" "./pages/ksphere/kubeflow"

# to make the parts more obvious:
npx fastmod --extensions md "\s*Regular_expression.*" "replacement" "./my/path"
```


.spelling
```diff
@@ -0,0 +1,147 @@
...
+SQLite
+KFServing
+MLOps
+MinIO
+MutatingWebhookConfiguration
+base64
+gRPC
+
+#
+# Stuff needed to make Kubeflow error free
```
samvantran, line 104: i like your idea of having an alphabetically sorted list so that it's easier to go through and fix. I personally like having files in subdirs for each product since that means they can all be sorted alphabetically but will leave it up to you to choose a global spelling file or not.

.spelling
```diff
@@ -0,0 +1,147 @@
...
+TFJob
+U.S.
+a.k.a.
+additionalMetricNames
+alice
+args
+autoscaling
+base-64-encode
+# consensus?
+concensus
```
samvantran, line 126: this is the wrong spelling and should be fixed with the line above.

.spelling
```diff
@@ -0,0 +1,147 @@
...
+containerd
+entrypoint
+funtions
```
samvantran, line 129: same here, this should be fixed in docs with `functions`

.spelling
```diff
@@ -0,0 +1,147 @@
...
+gRPC
+
+#
+# Stuff needed to make Kubeflow error free
+#
+28x28
+3rd-party
+ConfigMaps
+Eco-system
+# Should be ElasticSearch?
```
samvantran, line 110: does the spelling tool mark this correct when `Search` is capitalized or you'd have to add a rule for it in either case?

.spelling
```diff
@@ -0,0 +1,147 @@
...
+macOS
+mnist.py
+namespaced
+objectiveMetricName
+onboarding
+permissioned
+sigmoidal
+softmax
+stdout
+u
```
samvantran, line 139: this is not a spelling error?

.spelling
pierrebeitz: i'd actually leave that to @cmastr. i think that starting with a single global file brings most value as that enables us to spot/fix existing typos most easily.

i also think that having it completely sorted alphabetically is the best idea here. this PR is rather a demo of how things might work and that we build upon later.

.spelling
pierrebeitz: hunspell (the engine that's used under the hood of almost every open source checker) seems to go with smart-casing. so if we add `elasticsearch`, all things will be valid. but if we add `ElasticSearch`, then `Elasticsearch` will still be invalid.

.spelling
pierrebeitz: yeah, i was wondering about that one as well. i'm inclined to believe that this is either an entry in some table or a stray `u` that's used on some page (which is not a valid word obviously). 

the way to go forward would be to delete it from the dictionary and then use the interactive mode mentioned above to correct offences.

pierrebeitz: thanks for the review btw!

pierrebeitz:
after letting this sit for a day and receiving some feedback, i'm inclined to go all-in, meaning that we could add all offences to the global dictionary in a case-sensitive manner.

it would then be easy for people to (e.g. at a docathon) remove offences from that dictionary and fix typos in interactive mode afterwards. a pretty simple and effective workflow.

i'd consider that the method to bring value fastest (as it's quite easy to glance through an alphabetically sorted list and find bad things) while also being the method that allows us a strict CI-check right from the start.

i'd add convenience methods to invoke the report- and interactive-modes as well as adding all offences to the dict and document them properly.

what do you think?

samvantran:
Agree 💯 

> that allows us a strict CI-check right from the start.

liked this the most as it can stop future misspellings

.spelling
samvantran: probably we should not go with all lower since you don't want the tool to mistakenly allow spellings like `ElaSticsearch`

pierrebeitz:
superseded by #3047

PR #3045, opened 2020-06-24
makes markdown links that are in a different domain link out to a new…

msdolbey:
## Jira Ticket
N/A

So, currently, markdown has no out of the box solution for external links going into a new tab. However, that's the desired functionality that we want. So, I've been replacing different markdown links with a tags so that we can do that. 
But, as @samvantran and @dkoshkin have pointed out, that could lead to future issues when links in the future could turn to dead ones, that would be bad news bears and lead to customer issues. 

SO, we can code this solution for ourselves, and the linting/builds will still check to see if these links work.

This solution was found here: 
https://stackoverflow.com/a/4425214

## Checklist
- [x] Create your PR against `staging`, not `master`. 
- [ ] Provide an estimated date for deploying the doc change. Note: Improvements or fixes can be merged ASAP. 

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -2,3 +2,11 @@
 if (typeof feather !== "undefined") {
   feather.replace();
 }
+
+var links = document.links;
+
+for (var i = 0, linksLength = links.length; i < linksLength; i++) {
+   if (links[i].hostname != window.location.hostname) {
+       links[i].target = '_blank';
+   }
+}


js/main.js
```diff
@@ -2,3 +2,11 @@
 if (typeof feather !== "undefined") {
   feather.replace();
 }
+
+var links = document.links;
+
+for (var i = 0, linksLength = links.length; i < linksLength; i++) {
```
dkoshkin, line 8: Actually is it possible to limit these changes to the Ksphere docs? Other docs and teams might not want this behavior.

msdolbey: @cmastr - What do you think? ☝️

msdolbey: To follow up - Casie said let's make it universal.

PR #3096, opened 2020-07-17
Add make targets to build/run site in container, with "live editing"

dlipovetsky:
## Description of changes being made
Adds two make targets:
1. `make docker-liveedit-image` builds the site in a container
2. `make docker-liveedit` runs the site in the container, with live rendering of changes to files in the `pages` directory.

We would like to make use of this functionality to "live edit" konvoy docs, e..g,
```
make --debug docker-liveedit \
    LIVEEDIT_PAGES_SRC_ABS_PATH=/path/to/konvoy/repo/docs/site \
    LIVEEDIT_PAGES_DST_REL_PATH=ksphere/konvoy/latest \
    'LIVEEDIT_RENDER_PATH_PATTERN=ksphere/**'
```

The `docker-site-run` make target does run the site in a container, but requires docker-compose, and does not support live rendering.

This change does not affect any docs, or the existing toolchain.

## Checklist
- [x] Create your PR against `staging`, not `master`. 
- [x] Provide an estimated date for deploying the doc change. Note: Improvements or fixes can be merged ASAP. 

See the [contribution guidelines](https://github.com/mesosphere/dcos-docs-site/wiki/Contributing) for more information.

PR #3114, opened 2020-07-28
speed up live editing in docker …

pierrebeitz:
...and on bare metal.

1. the consistency-flag is mostly about optimizing performance on osx. we
trade off file consistency between the docker host and the container for
speed. we actually live dangerously here and say that the container is
authoritative (via "delegated"). this allows the container to perform
everything needed for a live reload way faster though and has never led
to a loss of information in past projects for me.

2. introducing yet another watch-rule for pages. the way the
watch-plugin is implemented limits file access in subsequent
metalsmith-runs to only those files that are on the right side of those
rules. because we need to access the `layouts`-folder whenever we change
a page, we formerly decided to just pump all markdown files
(`**/*.{md,tmpl}`) through the pipeline again. we now try to be more
clever and introduce a rule that always pumps the layout-files through
the pipeline whenever some page changed (unfortunately the plugin does
not seem to have a means to express that more intention-revealing).

  the main benefit here is that the live-reload server does not send ~6k
  events to the browser before that reloads anymore. a process which
  seems to be much slower in a docker-environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,9 +77,10 @@ docker-liveedit-image:
 
 docker-liveedit:
 	@test -n "$$(docker image ls --quiet $(LIVEEDIT_IMAGE))" || (echo "Image '$(LIVEEDIT_IMAGE)' not found. Did you already run 'make docker-liveedit-image'?"; exit 1)
+	# Note: --mount consistency=delegated for MacOS. See https://docs.docker.com/storage/bind-mounts/#configure-mount-consistency-for-macos.
 	docker run -it --rm \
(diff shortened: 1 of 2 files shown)


cmastr:
> 👍
> 
> shall we merge #3096 first and change head branch to staging or merge this into Daniel's PR?

Merged #3096. Base branch changed to Staging. Merge conflict in Makefile.

dlipovetsky:
I resolved the conflict (simple conflict--just the `consistency=delegated` option). I also added a comment to explain why the option is used.

PR #3123, opened 2020-07-30
remove obsolete entries from config.json

pierrebeitz:
there seem to be some obsolete branch-configs in `config.json`.

as i think it would be wise to completely get rid of this file in favour of a `draft`-mechanism, we now start by trimming it down.

1. `DO_NOT_INDEX` has no effect currently. things that are not built also won't be indexed.
2. the `develop`- and `alpha`-envs are not used anymore
3. `dispatch` does not have an according branch. IT MIGHT BE that someone specifies dispatch as a docker-arg in some build-process that is now slowed down. if so, they could use the RPP to only build dispatch, which would be even faster.
4. evade the need for `local` and `always` via changes in `index.js`.

--------------

this PR emerged while adding DCOS 2.2.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,30 +1,13 @@
 {
   "master": {
     "DO_NOT_BUILD": [
-      "test/**",
       "mesosphere/dcos/2.2/**"
-    ],
-    "DO_NOT_INDEX": [
     ]
   },
   "staging": {
     "DO_NOT_BUILD": [
-      "test/**",
       "mesosphere/dcos/2.2/**"
-    ],
-    "DO_NOT_INDEX": [
-      "^cn"
     ]
-  },
-  "develop": {
-    "DO_NOT_BUILD": [],
-    "DO_NOT_INDEX": [
-      "^cn"
-    ]
-  },
-  "local": {
-    "DO_NOT_BUILD": [],
-    "DO_NOT_INDEX": []
   },
   "kommander": {
     "DO_NOT_BUILD": [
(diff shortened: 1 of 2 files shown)


cmastr:
dispatch does not have an according branch. IT MIGHT BE that someone specifies dispatch as a docker-arg in some build-process that is now slowed down. if so, they could use the RPP to only build dispatch, which would be even faster.
 
Yeah this is @chhsia0 and the Dispatch team. I think there's some weirdness around it not building unless DCOS is involved so I'd like to have them take a look.

chhsia0:
Hi @pierrebeitz can you briefly describe what you mean by using RPP to only build dispatch? Right now we're passing `dsipatch` as branch name and relying on dispatch's `DO_NOT_BUILD` as you said, but I'm more than willing to use a faster approach :)

pierrebeitz:
Hi @chhsia0!

I just realized that the README does not contain all that much information about how to best run this (https://github.com/mesosphere/dcos-docs-site#previewing-your-work-and-formatting hints at it though).


My point is mainly about injecting a `RPP`-env variable into the context of `npm run dev`. E.g.
```
RPP="ksphere/dispatch/**" npm run dev
```

That will then tell metalsmith to (almost) only care about things within the dispatch-directory while also enabling a live-reload mechanism.

-------------

@dlipovetsky also just landed https://github.com/mesosphere/dcos-docs-site/pull/3096. With that you can edit (or preview) within docker conveniently.

PR #3124, opened 2020-07-30
enable incremental builds of swagger and ngindox

pierrebeitz:
this allows to quickly update those includes when adding a new one.
we formerly needed to delete all artifacts and recompile them, which
would take several minutes (like 5 on my machine).

we now evade the complexity of those scripts altogether and solve the
compilation in the Makefile. as an additional benefit we can now make
use of modern computers and compute all artifacts in parallel, which
IMMENSELY speeds up this process for me. that may lead to faster 
deployments as well.


scripts/ngindox.sh
```diff
@@ -1,87 +0,0 @@
...
-            clean_path=$(echo $d | sed 's/.*\.\/pages\///')
-            build_dir=$2/$clean_path/$file_basename
-            if [ -d "$build_dir" ]; then
-              rm -rf "$build_dir"
-            fi
-            mkdir -p $build_dir
-            printf "${GREEN}Temp Dir${BLUE} $build_dir${NC}\n"
-
-            # Run bootprint swagger-ui
-            touch "$build_dir/bundle.js"
```
pierrebeitz, line 63: the only difference is that we now do not create those empty `bundle.js` and `main.css` files anymore.

don't know what they were for, but the build seems to work fine without them.

PR #3135, opened 2020-08-05
add new chinese translations

pierrebeitz:
## Jira Ticket

https://jira.d2iq.com/browse/D2IQ-68500

Blindly copies over all files from the archive given in the linked JIRA.

* 🤔 this is about 200MB in size.
* 🤔 it changes the docs for old versions.

## to check this out locally

* pull this branch
* `make --jobs=$(nproc) build-api` (to build the new swagger-definitions
* `npm run dev`


cmastr:
1. not sure about the size! Is it bigger than the originals?
2. Yes we actually asked them to update 1.13 and 2.0 as well because they broke some links last time, plus we have made improvements that also need translation. So I'm glad they completed that!

pierrebeitz:
> not sure about the size! Is it bigger than the originals?

bigger than the previous chinese translations (which were like 8MB for each version). but not bigger than the respective english versions. apparently they added the image-folders (which pumps up the repo ~62MB per version of dcos). maybe we can leverage your have-a-default-and-override-as-needed-idea to also mitigate this behaviour later on. no action item here.

> Yes we actually asked them to update 1.13 and 2.0 as well because

👍

cmastr:
Local build seems to be good. Links are working (much better than before!) I think we can merge.

PR #3136, opened 2020-08-05
trim down algolia-plugin

pierrebeitz:
this is just a short cleanup.

we have a DO_NOT_INDEX-config that's not in use anymore. (except that we tell staging to not index `cn` - but guess what: we don't even index on staging).

thus we now could remove all the functionality that's present to figure out what NOT to index. please note that things in `DO_NOT_BUILD` also are not indexed - and i can currently not think of a use case where we want to deploy some pages but don't want to index them. and even if we ran into that case at some point in time, we could solve this a little more elegantly via a frontmatter-entry on respective pages/trees.

we also moved the algolia-env-vars to the plugin itself - as i feel they belong there and don't need to clutter our main-build-file.

and then we also hardcode the algolia projectId and public key. no
worries, we still can grep for all places. the ALGOLIA_INDEX was not in
use.


cmastr:
Is it possible that we do not want to index Chinese on production? I'm sure that would be done a different way anyway, but that's the only use case I can think of.

pierrebeitz:
don't know. we currently index it on production. and it'll be only a tiny change to not have it indexed, in case we ever want that.

pierrebeitz:
resolved conflicts

PR #3151, opened 2020-08-14
convert templates to use pugs class and id notation

pierrebeitz:
piped our layouts through a "pugifier".

```pug
# before
span(class="myClass")
div(class="myClass")


# after (div-elements are considered a default and are thus omitted)
span.myClass
.myClass
```

reasoning: less visual noise makes it easier to work with those
templates.


pierrebeitz:
resolved

PR #3158, opened 2020-08-18
D2iQ-41078

pierrebeitz:
this fixes the print-button.

due to some recent changes that add a `target="_blank"` to every
non-internal link, the print button broke.

you'd click it and it would navigate away to an empty page and then try
to print that empty page.

we now replaced the anchor with a button, as buttons don't suffer from
that target-changing we do.

i don't say printing is pretty. but at least we get to see something now again.


cmastr:
CDC5E48D-07AA-46BB-AAEF-D54F32842615_1_105_c

Looks MUCH better! And actually works now.

PR #3160, opened 2020-08-18
add vale

pierrebeitz:
along with the styles
* Microsoft
* [ttd-light](https://github.com/mesosphere/dcos-docs-site/pull/3160/files#diff-569b3ae9562e5b47d12016fe2e8fc4a3)
* write-good

please have a look at those styles. we may want to remove some of those rules right away or even add new ones.

we also add a Vocabulary named "[Docs](https://github.com/mesosphere/dcos-docs-site/pull/3160/files#diff-03e93b573cc056a71614e11183e5a0ab)".

note that there's one `accept.txt` and one `reject.txt`. we did not extensively test this, but vale claims that they also work with regular expressions. as a first test, see the reject rule for `on-premise[^s]`, which should tell us whenever someone tries to add `on-premise` to the codebase.

-------------------

This most likely will need further adjustment after merging.
Unfortunately there seems to be no easy way to test this github-workflow until
it's merged.


cmastr:
All signed off on this! Communication has gone out. Seem to be getting good feedback so far. Let's merge!

cmastr:
@pierrebeitz which of these files would be a good place to add our product names? I saw your note about reject.txt and accept.txt. < would that be a good place for product names?

pierrebeitz:
yes, that's exactly the place to put those!
as mentioned in the PR: vale also claims to accept regular expressions. you
could test your regular expressions at e.g. https://regexr.com/ or reach
out to me, if you need help with those.

PR #3173, opened 2020-08-23
restrict content-width

pierrebeitz:
on large monitors the main content can look a little off.
this tries to make it look a little better. maybe it's good enough to merge for now.

as no one asked for this, i'd be totally fine with
* closing this PR
* deciding that it looks good enough as it is or 
* starting a more involved process around it.

# before (scaled view)

before_small

# after

the content now appears centered in the middle, which gives the whole page more structure and prevents images from growing insanely wide.

after_small

### DIFF
--- a/scss/base/_content.scss
+++ b/scss/base/_content.scss
@@ -9,6 +9,8 @@ $sections-size: 20%;
     flex-direction: column;
     min-width: 0;
     padding: 40px;
+    max-width: 1000px;
+    margin: 0 auto;
   }
   &__404 {
     margin-bottom: -160px;


pierrebeitz:
☝️ @danielone WDYT?

pierrebeitz:
hmm... the margin would grow if we'd expand further, effectively limiting to 1000px width with this approach.

it is sometimes argued that sentences (and code) can be read better when the line length is limited. but i can totally see your point of using all space there is. as stated in the description i'm fine with closing this PR and will do so. if someone thinks this deserves more discussion they can reopen.

PR #3174, opened 2020-08-23
speed up deployment and setup

pierrebeitz:
I think this is a pretty impactful one. Here are the commit messages: 

## don't build dedicated swagger-docs

why? because they were only ever rendered to be included in PDFs.
so currently we're waiting like 10 minutes on every deployment (and on
every dev's machine that's setup to render docs) for no good reason.

if we want to build PDFs in the futue, we'd most likely do so by using
puppeteer or wkhtml2pdf, so even then this output is of no use.

## get rid of `make build-api` completely

we now compile ngindox-content on the fly, as the compiler seems to be
pretty darn fast. this way we can get rid of the whole "prebuild" step
and can start right at `npm run dev`. 

-----

imagine deploying or setting up a new dev-env in the future 😻 

# Testing

if you see that the includes work as they did before, everything should be fine!

* [prod ngindox](https://docs.d2iq.com/mesosphere/dcos/2.1/api/agent-routes ) / [local ngindox](http://localhost:3000/mesosphere/dcos/2.1/api/agent-routes ) (remember to checkout this branch and run `npm run dev` before clicking the local links)
* [prod swagger](https://docs.d2iq.com/mesosphere/dcos/2.1/gui/updating-gui/#updating-the-gui-via-the-api) / [local swagger](https://docs.d2iq.com/mesosphere/dcos/2.1/gui/updating-gui/#updating-the-gui-via-the-api)

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,12 +104,6 @@ Content that is kept on other team repos would follow their init, if applicable.
 This must be only done on first time or if a rare site tooling change occurs. Other repos may have other needs, see their Contributing Guide or similar resource for assistance if you will need to build their code.
 `npm install`
 
-### Build the API page sets
-This is necessary anytime a set is changed, or when a new DC/OS version (and there for a new set) is created
-`make --jobs=$(expr $(nproc) - 1) build-api`
-
-This takes some time. It builds the folders `build-ngindox` and `biuld-swagger`.
-
 ### Build a local preview
 This is to ensure setup was successful
 `npm run dev`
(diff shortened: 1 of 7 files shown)


cmastr:
YEAH! All for speedier builds.

README.md
```diff
@@ -357,6 +349,10 @@ Extra information that the user may wish to know, but not necessary to the basic
 ### Special - Adding API tables
 :TODO: add info on the API tables and the script to run
 
+### Removing a border from an image
+
+    ![Architectural overview](../img/Konvoy-arch-diagram.png){ data-no-border }
```
pierrebeitz, line 354: @cmastr i added 4 empty spaces here to make this render as a code block

PR #3189, opened 2020-08-28
allow to set RPP for `npm run build`

pierrebeitz:
we formerly made sure that an `RPP` could only be used in
development-mode. that restriction seems to be cumbersome for other
teams as they'd like to have a fast build (and not a fast dev-mode).
see
https://github.com/mesosphere/dispatch/pull/746#issuecomment-683142668

we have pretty solid control over how we deploy to production so it looks like it's worth removing this safety-measure in exchange for those (way!) faster builds.


gpaul:
Thanks!

PR #3214, opened 2020-09-11
Add Vale terms based on feedback from KUDO for Kubeflow

ianhellstrom:
Minimize false positives by adding more acceptable words and names.


.github/styles/Vocab/Docs/accept.txt
```diff
@@ -1,24 +1,64 @@
-D2iQ
-COPS
-Mesos
+addon
+Alertmanager
+Ansible
+autoscal(s|ing)
+Cassandra
+checkpoint(s|ed|ing)
```
porridge, line 6:
```suggestion
checkpoint(|s|ed|ing)
```
We want to allow `checkpoint` as well, right? Or is that accepted already?

porridge (approved):

.github/styles/Vocab/Docs/accept.txt
ianhellstrom: Good point!

ianhellstrom: I think that should already be a word on and of itself but the verb conjugations may not be. Still, I like making it clear that all versions are accepted.

.github/styles/Vocab/Docs/accept.txt
```diff
@@ -1,24 +1,67 @@
-D2iQ
-COPS
-Mesos
+addon
+Airflow
+Alertmanager
+Ansible
+autoscal(s|ed|ing)
```
cmastr, line 5: I don't think autoscal is a word

cmastr (changes requested):

.github/styles/Vocab/Docs/accept.txt
porridge: Pushed https://github.com/mesosphere/dcos-docs-site/pull/3214/commits/8068e2bfa9ed2e3b6d38baa1597e764942857d45 to fix.

cmastr (approved):

porridge:
Resolved the conflict and made sure the terms are sorted @cmastr

cmastr (approved):

PR #3225, opened 2020-09-18
Modify custom CA instructions for certificate renewal

jongiddy:
## Jira Ticket

https://jira.d2iq.com/browse/D2IQ-71528

Rewrite custom certificate page to add support for certificate renewal.

Although this will also work in the next minor versions of 2.1 and 2.0, I am just changing 2.2 for now to support the beta release. Once we're happy with 2.2 beta results and perhaps closer to minor updates to earlier versions, I will copy this back to 2.1 and 2.0

## Checklist
- [x] Test all commands and procedures, if applicable.
- [x] Create your PR against `staging`, not `master`. 

### DIFF
--- a/pages/mesosphere/dcos/2.2/security/ent/tls-ssl/ca-custom/index.md
+++ b/pages/mesosphere/dcos/2.2/security/ent/tls-ssl/ca-custom/index.md
@@ -11,7 +11,7 @@ model: /mesosphere/dcos/2.2/data.yml
 <!-- The source repository for this topic is https://github.com/dcos/dcos-docs-site -->
 
 
-Each DC/OS Enterprise cluster has a DC/OS certificate authority (CA). By default the DC/OS CA uses a globally unique root CA certificate generated during the installation of DC/OS. This CA certificate is used to sign certificates for the components of DC/OS such as Admin Router. Instead of using the auto-generated root CA certificate, you can configure DC/OS Enterprise to use a specific CA certificate, either a root CA certificate or an intermediate signing certificate. (See examples [below](#example-use-cases))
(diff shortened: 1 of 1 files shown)


cneth (changes requested):
Thanks for cleaning up the terminology on this page.  

However, I am not sure I understand how you 'replace' an existing cert.  Do you just reinstall like a 'patch' upgrade?   I assume there must be some restrictions on what you can change when you do this?   

It would be great if we had a 'worked example' for this procedure.

jongiddy:
@cneth Thanks for the comments.

Previously, we only permitted custom CA's during installation, and if they expired or were compromised they could only be changed by cluster reinstallation. Now we allow the CA certificate and key to be changed during installation, a version upgrade, or a patch upgrade. This changes the CA certificate on all nodes and replaces all the service certificates.

There's a few cases:
1. setup custom certificate during installation
2. add a custom certificate during a patch upgrade
3. replace an existing custom certificate with a new one during a patch upgrade ("cert renewal")
4. remove the custom certificate during a patch upgrade

I've currently just used one section for all 4 cases, but maybe I should separate them out.

From an implementation POV, case 2 is the same as case 3, but replacing the default CA with a custom cert. Case 4 is also the same as case 3 but replacing a custom cert with the default CA.

From a user perspective, case 1 and 2 both make the cluster use a CA selected by the user, for example, because the org wants to only use its certs. Case 3 provides "certificate renewal", although case 2 is also certificate renewal if you previously chose the default CA. Case 4 is an unlikely user requirement, so we could probably leave it undocumented (or mention it as an aside).

Do you think it would be clearer if I replaced the `Install or patch DC/OS` section with two sections?
- `Setting the CA certificate` (with a comment that not setting the attributes uses a default CA)
- `Renewing the CA certificate` (with a comment that unsetting the attributes reverts to the default CA)

cneth:
> Do you think it would be clearer if I replaced the `Install or patch DC/OS` section with two sections?
> 
> * `Setting the CA certificate` (with a comment that not setting the attributes uses a default CA)
> * `Renewing the CA certificate` (with a comment that unsetting the attributes reverts to the default CA)

Yes, I think this is a great idea, as those are the two main reasons you'd be referring to these docs.

cmastr:
going to start merging this. If any outstanding comments, open a new PR to address.

cmastr (approved):

cmastr (approved):

jongiddy:
Split into two sections described above will be done in https://jira.d2iq.com/browse/D2IQ-71896

PR #3239, opened 2020-09-23
`/ksphere` to `/dkp`

pierrebeitz:
steps taken:

* search for `docs.d2iq.com.*/ksphere/` and replace within the results.
* search for `("/ksphere/` and replace within the results. (usual markdown links)
* search for `: /ksphere/` and replace within the results. (reference-style links)
* search for `"ksphere/` and replace within the results. (some internal code)
* search for `"/ksphere/` and replace within the results. (some internal code)
* search for `'ksphere/` and replace within the results. (some internal code)
* search for `'/ksphere/` and replace within the results. (some internal code)
* manually adapt the redirects
* manually sieve through the remaining results and change where
appropriate.

there's a lot of space for errors here. i don't have any good advice on
how to review this. either we just try to go with it and you click
around a bit locally or you try to resemble something like the above
proceduce to see if you come up with the same result 🤷‍♂️

in any case we should test the redirections on staging once it's merged. 

# ⚠️ this PR might interfere with other PRs! 

those PRs might add files in `/ksphere`, which would need to be corrected after merging this. also they might introduce links to `/ksphere` which is also something we want to handle (mainly for SEO-reasons - which i don't know whether we really care about for docs).


docker/nginx/redirects-301.map
```diff
@@ -1,6 +1,7 @@
 ~^(.*/[^.^#^/]*[^/])$ $1/;
 ~^/1.7/.+ /1.7/;
 ~^/1.8/.+ /1.8/;
+~^/ksphere/(.*)$ /dkp/$1;
```
dkoshkin, line 4: Good idea.

dkoshkin (commented):
How can we verify that links still work.

pierrebeitz:
> How can we verify that links still work.

i'd suggest to first test this locally and see whether internal links still work (they seem to do as for how i tested it).

to see whether links that may occur on other sites still work, you might want to build a docker image containing those redirect files. you'd need to play around with `/docker/nginx/Dockerfile.production` a bit to make that work.
the other approach (the quick and dirty one) would be to merge this to staging and then test it there. in case of failure we just revert the commits. i'd call this more of a "pragmatic" (or quick-and-dirty) solution. but hey - no Docker-messing involved.

docker/nginx/redirects-301.map
samvantran: i figured all you'd need to do is rename the pages folder from `pages/ksphere` to `pages/dkp` and then add in this redirect. Wouldn't all the ksphere links redirect to dkp path based on this rule?

pierrebeitz: if we also changed the tooling-related things and the "later" redirect rules, that *would* work of course.
i think changing the links themselves is a good practice though and does not seem very risky. every client saves the redirect turnaround time and we can reason about how our internal links work more easily in a couple month from now when everyone has forgotten that this change happened. :)

i'd be open to compelling reasons not to change the links themselves though of course.

samvantran (commented):
Whenever I try to build a local site I run into 
```
/dcos-docs-site/index.js:326
  if (err) throw err;
           ^

Error: /dcos-docs-site/layouts/partials/sidebar.pug:7
    5| - var inMenu = function(c) { return c.menuWeight != -1 && (c.title || c.navigationTitle) }
    6| - var productPath = sphere == "services" ? `mesosphere/dcos${langFragment}/${sphere}/${product}` : `${sphere}/${product}${langFragment}`
  > 7| - var productVersions = hierarchy.findByPath(productPath).children.filter(item => semver.test(item.id) && item.menuWeight != -1)
    8| - var currentPath = '/' + path
    9|
    10| mixin renderSidebarNavListItem(item, depth)

Missing file in path: /ksphere/konvoy
    at Object.findByPath (/dcos-docs-site/plugins/metalsmith-hierarchy.js:93:13)
    at eval (eval at wrap (/dcos-docs-site/node_modules/pug-runtime/wrap.js:6:10), <anonymous>:634:33)
    at template (eval at wrap (/dcos-docs-site/node_modules/pug-runtime/wrap.js:6:10), <anonymous>:1188:146)
    at /dcos-docs-site/node_modules/consolidate/lib/consolidate.js:809:16
    at /dcos-docs-site/node_modules/consolidate/lib/consolidate.js:144:5
    at Promise._execute (/dcos-docs-site/node_modules/bluebird/js/release/debuggability.js:384:9)
    at Promise._resolveFromExecutor (/dcos-docs-site/node_modules/bluebird/js/release/promise.js:518:18)
    at new Promise (/dcos-docs-site/node_modules/bluebird/js/release/promise.js:103:10)
    at promisify (/dcos-docs-site/node_modules/consolidate/lib/consolidate.js:137:10)
    at exports.pug (/dcos-docs-site/node_modules/consolidate/lib/consolidate.js:793:10)
    at convert (/dcos-docs-site/node_modules/metalsmith-layouts/lib/index.js:149:7)
    at /dcos-docs-site/node_modules/async/lib/async.js:181:20
    at Object.async.forEachOf.async.eachOf (/dcos-docs-site/node_modules/async/lib/async.js:233:13)
    at async.forEach.async.each (/dcos-docs-site/node_modules/async/lib/async.js:209:22)
    at renderFiles (/dcos-docs-site/node_modules/metalsmith-layouts/lib/index.js:169:7)
    at /dcos-docs-site/node_modules/async/lib/async.js:52:16 {
  path: '/dcos-docs-site/layouts/partials/sidebar.pug'
}
```
I tried to fix all the paths pointing to `ksphere/konvoy` but still seem to run into the issue. That said, reviewing the changes looked mostly good. I'm for the practical approach of merging to staging and see what happens 🤞

pages/dkp/index.md
```diff
@@ -0,0 +1,10 @@
+---
+layout: konvoy-docs-landing.pug
+navigationTitle: D2iQ Kubernetes Platform
+title: D2iQ Kubernetes Platform
+featureMaturity:
+menus: ['landing-header']
+menuWeight: 5
+---
+
+Welcome to DKP
```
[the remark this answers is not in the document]
pierrebeitz: I believe that this dies not face customers anywhere at all. It MIGHT be that it ends up in our algolia (search-engine) index.

pierrebeitz:
> Whenever I try to build a local site I run into

@samvantran that error looks like it would appear whenever there's a page in `ksphere/konvoy` that popped up after the creation of this PR. The hierarchy-plugin will guess the "main" entrypoint for each product depending on the path. As `/pages/ksphere/konvoy/index.md` does not exist anymore, it guesses on something that does not exist anymore and thus crashes.

I think we should determine a point in time that we want to merge this, as it will be affected by any change (conflict) or addition (build breaks as above) of a file within `/pages/ksphere`. when we merged it to staging, i'd then pull it into all open PRs, so folks don't need to understand and wrestle with that change (as it likely causes broken builds or conflicts on their PRs).

any preferences on the timing or shall we just arbitrarily choose a day? @cmastr

cmastr:
> > Whenever I try to build a local site I run into
> 
> @samvantran that error looks like it would appear whenever there's a page in `ksphere/konvoy` that popped up after the creation of this PR. The hierarchy-plugin will guess the "main" entrypoint for each product depending on the path. As `/pages/ksphere/konvoy/index.md` does not exist anymore, it guesses a wrong one and thus crashes.
> 
> I think we should determine a point in time that we want to merge this, as it will be affected by any change (👉 conflict) or addition (👉 build breaks as above) of a file within `/pages/ksphere`. when we merged it to staging and that seems to work fine, i'd then pull staging into all open PRs, so folks don't need to understand and wrestle with that change (as it likely causes broken builds or conflicts on their PRs).
> 
> any preferences on the timing or shall we just arbitrarily choose a day? @cmastr

Timing wise, I think the earlier the better so we don't have massive conflicts like we do now. :(

cmastr:
Can we get rid of this one too?

BF57AF12-D83E-4613-953C-56F336D03939_4_5005_c

pierrebeitz:
> Can we get rid of this one too?

any thoughts on what to use there? i feel `D2iQ Kubernetes Platform` could be a bit long for that dropdown. `DKP` might be on the cryptic end though. 🤷‍♂️

samvantran:
> any thoughts on what to use there? i feel D2iQ Kubernetes Platform could be a bit long for that dropdown. DKP might be on the cryptic end though. 🤷‍♂️

For better or worse our marketing team has decided to go all in on DKP. Probably we should go with that but maybe `Kubernetes` is a better alternative.

cmastr:
> > Can we get rid of this one too?
> 
> any thoughts on what to use there? i feel `D2iQ Kubernetes Platform` could be a bit long for that dropdown. `DKP` might be on the cryptic end though. 🤷‍♂️

Just stick to DKP. I really don't like the switcher anyway, so if we want to do a bigger change, just get rid of this drop down.

cmastr:
Let's start over on this right after 1.6 Konvoy goes out. That way it's kind of clean slate for new versions going forward.

PR #3240, opened 2020-09-23
making links a bit more md friendly per sams feedback for kommander-i…

msdolbey:
…nstall-airgapped

## Jira Ticket
N/A

Per feedback from @samba - just making these internal links more markdown friendly, also having them be internal links.

## Checklist
- [ ] Create your PR against `staging`, not `master`. 

### DIFF
--- a/pages/ksphere/kommander/1.2/install-airgapped/index.md
+++ b/pages/ksphere/kommander/1.2/install-airgapped/index.md
@@ -29,19 +29,19 @@ Before installing, make sure your environment has the following basic requiremen
   - management cluster must connect to the attached cluster's API server
   - management cluster must connect to load balancers created by some addons. For example, Thanos, part of the Prometheus addon, connects to those load balancers.
 
-- all the prerequisites in [air-gapped Konvoy installation](https://docs.d2iq.com/ksphere/konvoy/1.6/install/install-airgapped/#before-you-begin) in case of Konvoy clusters.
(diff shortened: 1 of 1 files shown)


makkes (changes requested):
I'm not sure I understand the motivation here. My personal take is that I don't like this way of linking somewhere because as an editor I'd have to jump back and forth to see where the link goes to or to add new links. This makes link addition/removal/editing a tedious process, whereas the `[text](URL)` way of linking is self-contained and easily editable without having to jump back and forth in the document.

msdolbey:
> I'm not sure I understand the motivation here. My personal take is that I don't like this way of linking somewhere because as an editor I'd have to jump back and forth to see where the link goes to or to add new links. This makes link addition/removal/editing a tedious process, whereas the `[text](URL)` way of linking is self-contained and easily editable without having to jump back and forth in the document.

I'm a bit ambivalent to this as well, but I think it's something that we can try to be consistent on. Either way, if this doesn't get in or not, just responding to feedback. 

That being said, if we don't want to do it this way, we should probably keep the internal linky part. 

I'll let @russurquhart1 @cmastr decide what they would rather have.

russurquhart1 (commented):
I think this method is a more neat method, and is optimal if certain links get used a lot in a doc.

makkes (approved):
I'll bow to the majority.

cmastr:
I do like this model better. We are not using it anywhere but Konvoy atm. This is a good start, and we need to make sure to communicate that we are moving to this type of link structure in Kommander.

cmastr (approved):

PR #3233, opened 2020-09-22
D2IQ-71605 - [dcos] add new job-table fields to GUI docs

pierrebeitz:
closes D2IQ-71605

## Jira Ticket

https://jira.d2iq.com/browse/D2IQ-71605

## Description of changes being made

we added a couple new fields to the jobs table and now incorporate those in the docs.

## Checklist
- [x] Change all affected versions, if applicable (e.g. 1.13, 2.0, 2.1).
- [x] Test all commands and procedures, if applicable.
- [x] Create your PR against `staging`, not `master`. 
- [x] Update all links if you are moving a page.
- [?] Add release date to Release Notes page in the following format: <Package> was released on <Day>, <Month> <Year> ie `Mesosphere® DC/OS™ 2.1.0 was released on 9, June 2020`

See the [contribution guidelines](https://wiki.d2iq.com/display/ENG/Contributing+to+Docs) for more information.

## screenshot

Jobs - D2iQ Docs 2020-09-22 13-50-11

PR #3241, opened 2020-09-24
clean up the shortcodes file

pierrebeitz:
while working on #3233 i wanted to use the message-shortcode. it currently is not flexible enough for my use case. then these changes emerged before implementing what i need.

1. because they're not used in the codebase anymore we remove

  * `icon`
  * `switch`
  * `case`
  * `image`
  * `tooltip`

  validated via grepping for `"\[<SHORTCODE>"`.

2. remove `sanitize` as its name should rather had been `trim` and we
don't actually need it when writing our shortcodes differently.
3. add a helper for all those `badge` thingies. we might even introduce a `badge` shortcode that replaces all those later.
4. we remove the `fill` option from `message` as it's not used anywhere and even looks really broken currently.
5. we remove the "inline-block" `type` handling within the badge-helper as that's not used (anymore).
6. we remove `type` and `size` from `button` as those are not used anywhere.

i diffed the output before and after this change (with whitespace off) and did not notice any significant difference. maybe we can automate that in the future via minor adaptions to the sidebar rendering (which seems to spit out items in arbitrary order although they should be sorted) and the CSS (which uses `random` and thus produces different output with every build).

### DIFF
--- a/scss/shortcodes/_message.scss
+++ b/scss/shortcodes/_message.scss
@@ -45,31 +45,6 @@
     color: $color-dark;
     background: #89DA71;
   }
-  &--outline {
-    border: 1px solid $color-dark;
-    color: $color-dark;
-    background: transparent;
-  }
-  &--outline-warning {
-    border: 1px solid #F2D98E;
-    color: #F2D98E;
-    background: transparent;
-  }
-  &--outline-error {
-    border: 1px solid #E77975;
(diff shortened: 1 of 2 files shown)


samvantran (approved):
I might not be the best person to review but it lgtm

cmastr (approved):

PR #3292, opened 2020-10-22
Iterate on the deploy process and introduce preview environments

pierrebeitz:
## The boring bits

on the master pipeline we'll build and push a "builder" image to cache
npm dependencies on subsequent runs. actually that image should only
change, when we make adjustments to `package.json`.

this seems to save about 1.5 to 2 minutes on CI. 🎉

## And now for the cool stuff

We build a preview environment for each PR. If you like magical things,
you can insert your PRs number into this URL and put the result into
your browser:

`
docs-d2iq-com-pr-MY_PR_NUMBER.s3-website-us-west-2.amazonaws.com
`

If you like it comfy and have access to our jenkins, you can click on
the build result of `continuous-integration/jenkins/pr-merge`. Once that
is green its last output should give you exactly that link mentioned
above, but with the number already in it:

dcos-docs-site 2020-10-22 11-20-46

### Once it's green

Yes, we'll improve on that. Likely a bot will let you know of that URL in the future, so you don't have to browse Jenkins or put your PR-number into the URL above. But hey! We now have preview environments for the first time in docs history!

Yes, we don't clean up those environments yet, but as each env should
cost about 2 cents a month, it feels like we have a couple more days to
come up with a cleanup-routine. 🤞


cmastr (approved):

cmastr:
Woot! This is awesome sauce.

PR #3360, opened 2020-11-24
remove addThis

pierrebeitz:
addThis is a service that makes for a great share-button.
unfortunately they add 9 script files in total that weight in at ~1MB
and they add a click handler that catches clicks on internal and
external links which makes for a noticeable delay in changing pages.

as we guess that this button is not used very often, we remove it for
now. if we ever want to have share functionality, that's pretty easy
(and also stable by now) to add for the major platforms even without
using a service.

--------

obviously today is docs-performance-day...

### DIFF
--- a/js/actions.js
+++ b/js/actions.js
@@ -1,8 +1,0 @@
-window.addEventListener("load", () => {
-  if (window.addthis) {
-    window.addthis_config = {
-      ui_disable: true,
-    };
-    addthis.init();
-  }
-});
--- a/js/main.js
+++ b/js/main.js
@@ -10,7 +10,6 @@ require("./sidebar-drawer.js");
 require("./header.js");
 require("./header-dropdown.js");
 require("./search.js");
-require("./actions.js");
 require("./sections.js");
 require("./switch.js");
 require("./image.js");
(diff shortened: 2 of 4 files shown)


cneth (approved):

cmastr (approved):

msdolbey:
Had one weird experience - when I ran `npm install`, it resulted in an error. BUT, I switched branches to see if that would build, then switched back, and it was fine. So, it was likely just a my machine problem. 

Other than that, looks good to me!

msdolbey (approved):
👍 
Remove This!

pierrebeitz:
@msdolbey as long as the CI-check with the beautiful name `continuous-integration/jenkins/pr-merge` is green, CI is able to `npm install` and `npm run build`. additionally we did not touch related code in this PR which why i'm very sure it's safe to merge.

i have a hunch that the error you saw **might** be related to a recent change that makes our builds repeatable - in that the exact dependency resolution is now in the repo.

if you should ever encounter something like this again, i'd be happy to figure out the root cause if you share a stack tract or screenshot or put up a zoom in the morning hours.

PR #3363, opened 2020-12-02
decrease build size

pierrebeitz:
a lot of the markup on our pages belong to the sidebar, as we always
render all entries and each of those has really long classnames attached
to it.

`du -sm` tells us `1482MB` before and `903MB` after this change.
if we take into account that images weight in at `749MB` (and thus are our next
target), we get a html-size reduction of about 80% (from `733MB` to `154MB`).   
the size of e.g. `mesosphere/dcos/2.2/release-notes/2.2.1/` went from 160KB to 15KB.

benefits: 

* upload to S3 is faster (we should see deployments that scratch the 4 minute mark regularly with this)
* metalsmith's layouts-step takes less time (faster deployment, faster dev-cycle)
* ~40% less spend on S3 
* potentially better SEO-score due to improved load times

the change itself is twofold:

1. don't render nested children that would be invisible
2. don't use BEM for sidebar-CSS as the long class names are the culprit here.

## testing

* it's mainly about the sidebar-navigation or "TOC". so if that still works as expected, we're good.

⚠️ TRADEOFF: we can't click the carets anymore to dynamically open a navigation item with children on the same page. instead the associated site will open (with the now open sub-menu). if you think that drawback in UX outweighs the gains described above, please let me know.

----------------

btw sorry for putting up a non-related PR when we have issues in JIRA to be solved - but this kind of tiny and straightforward improvement fits into a few waiting cycles i encounter in the migration-work i'm currently doing while not requiring much overhead due to a context switch.

### DIFF
--- a/js/sidebar.js
+++ b/js/sidebar.js
@@ -1,24 +1,11 @@
 const sidebarDropdown = document.querySelector(".sidebar__dropdown");
-const sidebarDropdownList = document.querySelector(".sidebar__dropdown__list");
-let sidebarItems = document.querySelectorAll(".sidebar__nav__item--parent");
-sidebarItems = [...sidebarItems];
+const sidebarDropdownList = document.querySelector(".sidebar__dropdown ul");
 
 if (sidebarDropdown && sidebarDropdownList) {
   document.addEventListener("click", detectClick);
   sidebarDropdown.addEventListener("click", toggleDropdown);
 }
 
-if (sidebarItems.length) {
-  sidebarItems.forEach((menu) => {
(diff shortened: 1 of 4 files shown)


cmastr:
I think this seems ok. It seems very similar to the change already made with the TOC.

cmastr (approved):

samvantran:
I think this is fine but maybe we should've consulted with Support as they heavily use the docs site

That said, 40% less spend on S3 and 80% reduction in file size - hard to argue

PR #3407, opened 2021-01-27
add a page that lists CVEs in an interactive table

pierrebeitz:
this adds a page (`/dkp/security-updates/` until someone has a better suggestion) to show the results of our CVE-scan.

currently the file is committed to this repo, as we don't have an automated export to some S3-bucket yet. once that's in place we only need to change the URL we fetch though. 🎉 

i chose to already implement some sorting and pagination as without that the page felt kinda crappy. it's still not looking great, but i think it should be good enough to release it.

 
screencast 2021-01-27 16-59-31


mhrabovcin:
@pierrebeitz its not clear from the demo if this is sorting only per page.

pierrebeitz:
we sort the whole dataset before paginating, as i'd expect that as a user wanting to systematically browse through the data. 
would you expect only the current page to be sorted or just want to make sure we're shipping a proper implementation?

mhrabovcin:
> we sort the whole dataset before paginating, as i'd expect that as a user wanting to systematically browse through the data.
> would you expect only the current page to be sorted or just want to make sure we're shipping the "right" thing?

Thats the behavior I'd expect. It wasn't clear to me from the gif demo if that is the case. Thank you!

pages/dkp/security-updates/index.md
```diff
@@ -1,8 +1,8 @@
 ---
 layout: layout.pug
 navigationTitle: Security Updates
-title: Secutiry Updates
```
pierrebeitz: ouch. thanks!

cmastr (approved):

pierrebeitz:
@tillt, @mhrabovcin you might want to have a look here: http://docs-d2iq-com-pr-3407.s3-website-us-west-2.amazonaws.com/dkp/security-updates/. 

let me know if we can improve on anything!

cmastr:
Holding this for now until we have final signoff around the formatting and annotations.

russurquhart1 (approved):

PR #3440, opened 2021-02-19
mark mitigated entries and add an according filter

pierrebeitz:
this lets customers see the difference between mitigated and non-mitigated issues.

things done and looking for more opiniones/suggestions:

- grey out the background of a mitigated CVE
- strikethrough the severity when the CVE is mitigated
- add a tooltip to this striked severity: **"Mitigated - see Addtional Explanation"**
- add a filter-button right to the severity-filters to hide mitigated CVEs. it's filtering by default, so one has to click it to see all the CVEs of the export.
- move filters to the left, so we can fit one more button easily above the table
- adapted responsiveness (so it does not look rubbish on mobile devices)

we might want to run this by the UX-department as i feel like that
eye-button is rather not intention-revealing at first sight.

# have a look here:

https://user-images.githubusercontent.com/300861/108495520-2cb48680-72a9-11eb-88fc-91646c1052c6.mp4

### DIFF
--- a/js/cve.js
+++ b/js/cve.js
@@ -12,6 +12,7 @@ import React from "react";
 ///////////////////////////////////////////////////////////////////////////////
 //                                   MODEL                                   //
 ///////////////////////////////////////////////////////////////////////////////
+const perPageOptions = [10, 25, 50, 100, 200];
 const init = {
   // all cves we got from the server.
   cves: null,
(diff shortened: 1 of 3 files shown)


cmastr (approved):

cmastr:
Yeah I like this approach! Makes it more clear and allows filtering.

PR #3444, opened 2021-02-23
clean up images

pierrebeitz:
TL;DR
----------------
the build size is now down from almost a gigabyte to less than ~470MB.

i ran a script to search and delete images that are not referenced anymore. examining a couple of those by hand it looks like they're mostly DC/OS-images from ancient versions.

then i ran a resize-script on all images as it's rather untypical to ship a 4000x3000 image with a couple MB on the web. i arbitrarily chose a maximum of 1600px in width (we can talk about that if anybody thinks that 1600 in either height of width is too low quality - remember: it's only about images that had larger dimensions than this - which were only a few). 

i also ran `optipng` with an optimization level of 8 over night on all pngs (no quality loss) to get an overall reduction of about 22% on the remaining files. 

**benefits?**

* less spend on S3 (about 50% reduction)
* faster page loading for users
* less load and bandwidth-usage on CI (note that the sheer size of the build was the reason we had to remove DC/OS-versions from the repo and introduce an archive-docker-image - we can now feel freed from that restriction and more flexible for the future again)


cmastr:
i ran a script to search and delete images that are not referenced anymore. examining a couple of those by hand it looks like they're mostly DC/OS-images from ancient versions.

NICE!!

then i ran a resize-script on all images as it's rather untypical to ship a 4000x3000 image with a couple MB on the web. i arbitrarily chose a maximum of 1600px in width (we can talk about that if anybody thinks that 1600 in either height of width is too low quality - remember: it's only about images that had larger dimensions than this - which were only a few).

This is definitely better than before. I WOULD like to see a percentage when it comes to images. So that in case someone is using a tablet or a phone to view docs, that the image is resized in relation to the screen, but with a maximum size. ;)

cmastr (approved):

cmastr:
Do these changes need to be backported to the respective repos? Konvoy, Dispatch, Kaptain?

pierrebeitz:
> Do these changes need to be backported to the respective repos? Konvoy, Dispatch, Kaptain?

thanks for thinking of this! yes, we need those!

> I WOULD like to see a percentage when it comes to images. So that in case someone is using a tablet or a phone to view docs, that the image is resized in relation to the screen, but with a maximum size. ;)

as we have responsive images (right?), i think you're aiming at sending smaller images to those clients, right? while that's a good practice i don't see a big benefit of doing so. usually you want to do that to improve the loading times on image-heavy sites (which i've not yet seen any on the docs) in order to have a better user experience and most importantly a better SEO-rating. not sure if that's worth doing for us right now.

cmastr:
I agree on the "right now" part

cmastr:
Do these changes need to be backported to the respective repos? Konvoy, Dispatch, Kaptain?

thanks for thinking of this! yes, we need those!

On this part, I'm going to hold merging this until we can talk to the teams about getting these changes into their repos.

pierrebeitz:
superseded by #3462

PR #3456, opened 2021-02-28
Clean up frontmatters, add handlebars-templating and fix some bugs

pierrebeitz:
in this episode of "the big docs-cleanup" we introduce a working handlebars-setup and replace the "include-plugin". we also learned that the data-loader-plugin is just injecting variables for potential use in mustache/handlebars and that we can interpolate arbitrary frontmatter-entries. e.g.:

```
---
title: blorgh
some:
  nested: stuff
---

# DKP-{{ some.nested }} is awesome!
```

also we now force every markdown-file through the layout-step and thus discovered some issues on production: 
* some pages have not been rendered due to invalid frontmatter.
* some pages have non-correct code-snippets as our "in-place-dcos-plugin" just removed stuff in `{{ }}` sometimes. we now should get errors whenever someone does not escape the curlies like so: `\{{ }}` because we propagate the resulting error up the chain and won't compile that page. (unless you're really unfortunatel and write a valid templating expression like `{{ title }}` or `{{ except }}` - but that is likely catched when having a look at the page in the preview-env). Note that we also escaped 

---------------

please have a look at the separate commits to review this. i'm pretty sure that nothing broke, because as we now have repeatable builds, we can `diff -r` on a build before these and after these changes to see the difference. all of the difference should be fixing bugs.

please review `A short intro` here: https://github.com/mesosphere/dcos-docs-site/blob/pb/chores4/README.md#a-short-intro. it's meant to bring folks up to speed concisely as the former readme was way too long for my attention span :D

⚠️  this work would need backports to the according repos again, so maybe we merge this along with the image-compression-effort.


README.md
```diff
@@ -36,8 +36,71 @@ npm run dev
...
+
+- **excerpt**: presented below the title at the top of the content page, sometimes used in topic cards
+- **layout**: this determines the template you need, defaults to `layout.pug` (the standard content-page-layout).
+- **menuWeight**: determines ordering in the navigation structure, all numerical values except -1 accepted and sorted.
+- **navigationTitle**: short title that shows up in navigation bar on the left.
+- **title**: shows up at top of the page
+
+----------
+
+* **beta: true** - adds an beta label to a page's titel
```
russurquhart1, line 55:
```suggestion
* **beta: true** - When this value is **true** a beta label is added to the page title.
```

README.md
```diff
@@ -36,8 +36,71 @@ npm run dev
...
+* **experimental: true** - adds an experimental label to a page's titel
```
russurquhart1, line 56:
```suggestion
* **experimental: true** - When this value is **true** an experimental label is added to the page title.
```

README.md
```diff
@@ -36,8 +36,71 @@ npm run dev
...
+
+## URL-structure
+
+The directory-tree withing `/pages` resembles the URL-structure of the final build. A pages' content **MUST** be in an `index.md`-file within the according directory.
```
russurquhart1, line 60:
```suggestion
The directory-tree in `/pages` resembles the URL-structure of the final build. A pages' content **MUST** be in an `index.md`file in the respective directory.
```

README.md
```diff
@@ -36,8 +36,71 @@ npm run dev
...
+
+In short: `/pages/some/directory/index.md` will be available at `docs.d2iq.com/some/directory`.
```
russurquhart1, line 62:
```suggestion
For example, `/pages/some/directory/index.md` is located at `docs.d2iq.com/some/directory`.
```

russurquhart1 (approved):

cmastr:
The only thing I'm worried about with these changes is making sure we duplicate anything changed into their respective repos. For example, are any of these variables impacted in Konvoy docs? If so, let's make sure we backport those changes. (Same as the images changes)

cmastr (approved):

pierrebeitz:
will give this a second pass soon in a new PR.

PR #3543, opened 2021-04-22
chore: consolidate mustache-tokens

pierrebeitz:
let's have them all written in the same way in order to have an easier
time detecting them in the future programmatically.


pierrebeitz:
i made sure that this does not change the build programmatically. thus i'm merging.

to be clear: the build is changed a tiny bit as there are pages where our mustache-variables are not replaced! No worries, we're about to fix that!

Diagnostic Tools - D2iQ Docs 2021-04-22 13-41-25

PR #3544, opened 2021-04-22
chore: remove obsolete frontmatter-entries

pierrebeitz:
removing: 
* `category: K-Sphere` (has no meaning to the build-process)
* `featureMaturity` (has no meaning to the build-process)


pierrebeitz:
merging as the build-output is exactly the same as before. we can revert if you think that one of those entries has some value.

PR #3627, opened 2021-05-27
fix: use --delete option when rsync from upstream

dkoshkin:
Avoid issues like https://github.com/mesosphere/dcos-docs-site/pull/3626/files

```
--delete                delete extraneous files from destination dirs
```

Im assuming here that the upstream repo will be the only source of truth and all files must come from there.


dkoshkin:
Tested by running it locally and it generated this commit https://github.com/mesosphere/dcos-docs-site/commit/37b9826977b3a21f0e725b91b4eca33beb99a041

samvantran (approved):

PR #3631, opened 2021-06-03
Deployment changes for D2IQ-75904 and D2IQ-75903

lloesche:
This should implement the changes requested in D2IQ-75903 and D2IQ-75904.
The `--delete` flag to `aws s3 sync` makes sure that files that do not exist in the local `build/` directory are being removed from the remote bucket.
The new `beta` branch will be synced to the existing staging bucket. Once the Jenkins job is created and runs against the `beta` branch we'll check if the `docs-d2iq-com-staging` bucket was synced successfully and point the https://beta-docs.d2iq.com/ Cloudfront distribution to the staging bucket instead of the PR 3578 bucket.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,9 +1,17 @@
 #!/usr/bin/env groovy
 
 boolean main = env.BRANCH_NAME == "main"
-def bucket   = main ? "production"    : "pr-${env.CHANGE_ID}"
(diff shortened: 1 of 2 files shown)


lloesche:
@fatz can you please review if the Jenkinsfile syntax is correct like this? I've never written a Jenkinsfile and just adapted the existing one.

fatz (approved):

PR #3714, opened 2021-08-12
feat: combine konvoy and kommander into one pr

mstruebing:
## Jira Ticket
Before creating this pull request, make sure you have a separate ticket for the docs team to track their work. If you need assistance, ask in the #documentation Slack channel.

https://jira.d2iq.com/browse/D2IQ-77810

This combines kommander2/konvoy2 PRs into one sync PR.
I left the single ones alive as a fallback for now, until we have tested this a bit.

https://mesosphere.slack.com/archives/C02AVBS638R/p1628286710018400

## Checklist
- [ ] Create your PR against `main`. 

### DIFF
--- a/ci/pull-upstreams.sh
+++ b/ci/pull-upstreams.sh
@@ -1,0 +1,109 @@
+#!/bin/bash
+
+# Requires git >= 2.19 for sparse checkout
+# Requires rsync
+
+set -euxo pipefail
+IFS=$'\n\t'
+
+PROJECT_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")"/../ > /dev/null 2>&1 && pwd)"
+# Directory where the other repositories are cloned into
+SYNC_DIR="$PROJECT_ROOT/sync"
+mkdir -p "$SYNC_DIR"
+
+# Manipulates global BRANCH_NAME
+function construct_branch_name() {
+  local ARGC="$#"
+
+  while [ "$ARGC" -ge 4 ]; do
+    if [ "$ARGC" -eq 4 ]; then
+      BRANCH_NAME="$BRANCH_NAME$1/$3"
+    else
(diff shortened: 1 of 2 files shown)


ci/pull-upstreams.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
```
vacas5, line 1: what's different now that this file needs to be committed?

vacas5 (commented):
I may need a bit to wrap my head around this. I understand what's happening in the cron job. Does this effectively accomplish having a single PR that we could potentially proxy behind dev-docs.d2iq.com as mentioned?

ci/pull-upstreams.sh
```diff
@@ -0,0 +1,109 @@
...
+}
+
+# Will be manipulated by construct_branch_name and construct_pr_name
+PR_NAME="Sync"
+BRANCH_NAME="autosync/"
+
+# Renamed for better readability
+ARGC="$#"
+
+construct_branch_name "$@"
```
vacas5, line 86: I wonder if we should simply create a long lived branch. I guess the benefit here is that we don't have to keep rebasing it with changes on main.

ci/pull-upstreams.sh
mstruebing: This file can pull more than one other repo into one PR while the other only is able to pull one.

mstruebing:
> Does this effectively accomplish having a single PR that we could potentially proxy behind dev-docs.d2iq.com as mentioned?

It has a single branch, but I think the intention here (at least from my side) was to be easier to be reviewed from docs people.
We may use it also for proxying, but I didn't thought about it yet as I wasn't aware that this is a combined requirement.

vacas5 (approved):
I think this is an improvement! thanks @mstruebing

takirala (approved):
:tada:

PR #3931, opened 2021-12-03
D2IQ-81851: vale + remark

mstruebing:
## Jira Ticket
Before creating this pull request, make sure you have a separate ticket for the docs team to track their work. If you need assistance, ask in the #documentation Slack channel.

## Checklist
- [ ] Create any new PRs against `production`. 


clintonmedbery (approved):
Nice!

cmastr:
Tell me more about what this does?

clintonmedbery:
Basically adds linters, link checkers etc to a precommit hook. So you make some changes, try to commit and then your computer yells at you about what you did wrong.

vacas5:
@cmastr , this is adding in the link checking that was requested, as well as running vale locally before a developer makes a commit, so that they can see what they need to change and change it at the time rather than after pushing and getting frustrated.

.husky/pre-commit
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env sh
+. "$(dirname "$0")/_/husky.sh"
+
+if ! command -v vale &> /dev/null; then
```
vacas5, line 4: tell me more about this. Does it not come installed with NPM?

clintonmedbery: Vale is not an npm package.

vacas5 (changes requested):
I'd like relevant readme changes to be a part of this PR

vacas5:
@cmastr we need some direction here as to what to do with Vale. We'll have to have users install it on their local machines as well as all the JS stuff to use.

However, I think it might be really worth while. These tools basically act as a first pass code review. Vale will make suggestions on changes that will save the docs team time if developers or the community are making text suggestions.

cmastr:
> @cmastr we need some direction here as to what to do with Vale. We'll have to have users install it on their local machines as well as all the JS stuff to use.
> 
> However, I think it might be really worth while. These tools basically act as a first pass code review. Vale will make suggestions on changes that will save the docs team time if developers or the community are making text suggestions.

I am ok with requiring them to install it locally. I think that would actually be preferred for all. I think we will need clear instructions for that and we can help write those.

vacas5:
@clintonmedbery for `markdownlint` did you pull over the config from Kommander/Konvoy or not yet?

clintonmedbery:
I thought there was not one since I didn't see one in kommander but apparently it points to https://github.com/mesosphere/dkp-docs-tooling/blob/26061c69a6b8796a95617ef152539c05c8a77e58/.markdownlintrc in a submodule.
So no.

vacas5 (approved):

PR #3978, opened 2021-12-17
redirect for dispatch broken link

cmastr:
## Checklist

### DIFF
--- a/s3config.json
+++ b/s3config.json
@@ -53,8 +53,7 @@
     { "Condition": { "KeyPrefixEquals": "services/" },                                                         "Redirect": { "HostName": "$REDIR_HOSTNAME", "ReplaceKeyPrefixWith": "mesosphere/dcos/services/", "HttpRedirectCode": "301" } },
     { "Condition": { "KeyPrefixEquals": "version-policy/" },                                                   "Redirect": { "HostName": "$REDIR_HOSTNAME", "ReplaceKeyPrefixWith": "mesosphere/dcos/version-policy/", "HttpRedirectCode": "301" } },
     { "Condition": { "KeyPrefixEquals": "support" },                                                           "Redirect": { "HostName": "support.d2iq.com", "Protocol": "https", "HttpRedirectCode": "301" } },
-    { "Condition": { "KeyPrefixEquals": "dkp/kommander/1.4/Architecture/" },                                "Redirect": { "HostName": "$REDIR_HOSTNAME", "ReplaceKeyPrefixWith": "dkp/kommander/1.4/architecture/", "HttpRedirectCode": "301" } }
-    { "Condition": { "KeyPrefixEquals": "ksphere/dispatch/" },                                              "Redirect": { "HostName": "$REDIR_HOSTNAME", "ReplaceKeyPrefixWith": "dkp/dispatch/", "HttpRedirectCode": "307" } },
-
+    { "Condition": { "KeyPrefixEquals": "dkp/kommander/1.4/Architecture/" },                                "Redirect": { "HostName": "$REDIR_HOSTNAME", "ReplaceKeyPrefixWith": "dkp/kommander/1.4/architecture/", "HttpRedirectCode": "301" } },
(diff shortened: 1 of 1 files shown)


msdolbey:
Making a quick change - standby, @cmastr

fatz (dismissed):
I'm pretty sure these rules get applied in order so I would suggest moving this new rule before line 45 https://github.com/mesosphere/dcos-docs-site/blob/64178ebfcc7726027142f57b475c32427148ab67/s3config.json#L45

msdolbey:
Oh, that's a good point - @fatz - 

Actually, because of that - this may be unnecessary - /ksphere/dispatch should already re-direct to /dkp/dispatch.

Standby.

msdolbey:
We're good already here - so it will be weird locally no matter what, and seemingly on dev, but that is the same for dispatch as it is for konvoy/kommander, etc.

But, for the real show, it's good on the docs site as is already.

FYI @cmastr - I think we can close this.
https://user-images.githubusercontent.com/5703649/146585855-e14348c7-3acb-4062-b79d-824d1a96c013.mov

cmastr:
oh weird. Why is it not working for @gflorinskaya then??

cneth:
Hey @msdolbey - not sure why you say these are working...   I just tried it this link:

https://docs.d2iq.com/ksphere/dispatch/latest/

And it returns a not found.  However, we can see what the problem is too, it's redirecting to here:

http://docs.d2iq.com/dkp/dispatch/1.2/

@cmastr The problem is this rule:

{ "Condition": { "KeyPrefixEquals": "dkp/dispatch/latest/" }, "Redirect": { "HostName": "$REDIR_HOSTNAME", "ReplaceKeyPrefixWith": "dkp/dispatch/1.2/", "HttpRedirectCode": "307" } },

We archived 1.2 so this is why you get the dead links.   This should be 1.4 here.     I think if you make just this change and not the one you added, it will work.

msdolbey:
Oh - I see @gflorinskaya comment - it's because dispatch's 'latest' was never updated - lemme push something.

msdolbey:
@cmastr - ready to roll
@cneth - check too if you have some time.

cneth (approved):
This is the right fix.   Here's the current sequence of redirects:

$ curl -L -v https://docs.d2iq.com/ksphere/dispatch/latest/  2>&1 >/dev/null | grep -e GET -e location:
> GET /ksphere/dispatch/latest/ HTTP/2
< location: http://docs.d2iq.com/dkp/dispatch/latest/
> GET /dkp/dispatch/latest/ HTTP/1.1
> GET /dkp/dispatch/latest/ HTTP/2
< location: http://docs.d2iq.com/dkp/dispatch/1.2/
> GET /dkp/dispatch/1.2/ HTTP/1.1
> GET /dkp/dispatch/1.2/ HTTP/2

The ksphere -> dkp happens on the first redirect, then the 'latest' -> '2.1' on the second round, which is what fails.   So this will solve it.

msdolbey:
> This is the right fix. Here's the current sequence of redirects:
> 
> $ curl -L -v https://docs.d2iq.com/ksphere/dispatch/latest/ 2>&1 >/dev/null | grep -e GET -e location:
> 
> > GET /ksphere/dispatch/latest/ HTTP/2
> > < location: http://docs.d2iq.com/dkp/dispatch/latest/
> > GET /dkp/dispatch/latest/ HTTP/1.1
> > GET /dkp/dispatch/latest/ HTTP/2
> > < location: http://docs.d2iq.com/dkp/dispatch/1.2/
> > GET /dkp/dispatch/1.2/ HTTP/1.1
> > GET /dkp/dispatch/1.2/ HTTP/2
> 
> The ksphere -> dkp happens on the first redirect, then the 'latest' -> '2.1' on the second round, which is what fails. So this will solve it.

Wow - you went deeper than I ever did on this. Thanks for showing the work on my proof set, @cneth

gflorinskaya (approved):
Looks good to me, https://docs.d2iq.com/dkp/dispatch/1.4/ should work

PR #4015, opened 2022-01-06
fix kommander airgapped instructions

takirala:
## Checklist

### DIFF
--- a/pages/dkp/kommander/2.1/install/air-gapped/index.md
+++ b/pages/dkp/kommander/2.1/install/air-gapped/index.md
@@ -216,7 +216,7 @@ Based on the network latency between the environment of script execution and the
 1. Download and extract the `kommander-applications` bundle.
 
     ```bash
-    mkdir kommander-applications && wget https://github.com/mesosphere/kommander-applications/archive/refs/tags/${VERSION}.tar.gz -O - | tar xvzf - -C kommander-applications --strip-components 1
+    mkdir kommander-applications && wget https://downloads.mesosphere.com/dkp/kommander-applications_${VERSION}.tar.gz -O - | tar xvzf - -C kommander-applications --strip-components 1
     ```
 
 1. To install Kommander in your air-gapped environment using the above configuration file, enter the following command:
--- a/pages/dkp/kommander/2.1/install/air-gapped/index.md
+++ b/pages/dkp/kommander/2.1/install/air-gapped/index.md
@@ -213,10 +213,16 @@ Based on the network latency between the environment of script execution and the
       helmMirrorImageTag: "${VERSION}-amd64"
     ```
 
+1. Download and extract the `kommander-applications` bundle.
+
+    ```bash
+    mkdir kommander-applications && wget https://github.com/mesosphere/kommander-applications/archive/refs/tags/${VERSION}.tar.gz -O - | tar xvzf - -C kommander-applications --strip-components 1
+    ```
+
 1. To install Kommander in your air-gapped environment using the above configuration file, enter the following command:
 
     ```bash
-    kommander install --installer-config ./install.yaml
+    kommander install --installer-config ./install.yaml --kommander-applications-repository ./kommander-applications
     ```
 
 1. If you are installing Kommander in an AWS VPC, set the Traefik annotation to create an internal facing ELB by creating the following configmap in the `kommander` namespace:


pages/dkp/kommander/2.1/install/air-gapped/index.md
```diff
@@ -213,10 +213,16 @@ Based on the network latency between the environment of script execution and the
       helmMirrorImageTag: "${VERSION}-amd64"
     ```
 
+1. Download and extract the `kommander-applications` bundle.
+
+    ```bash
+    mkdir kommander-applications && wget https://github.com/mesosphere/kommander-applications/archive/refs/tags/${VERSION}.tar.gz -O - | tar xvzf - -C kommander-applications --strip-components 1
```
gracedo, line 219: Would we need to use our hosted bundle at downloads.mesosphere.com instead for any particular reason? cc @cneth if you know

some-things: We wouldn't need to, as long as everything stays consistent with the docs. The only thing that I could see is it could cause some confusion having the bundle on the site at downloads.mesosphere.com, and the docs reference the GitHub links.

msdolbey: I know this is also out there right now - but for the vanilla Kommander - we point to the downloads page. Should we do something similar for this? This seems pretty exposed for anyone who wants it right now.

takirala: While i like the fact that everything coming from downloads.mesosphere.io bucket would contribute better UX, in this particular context i think we want to "advertise"/"educate" users on how/where the repo is built or comes from.

We also use a same style of url for the catalog github repository https://github.com/mesosphere/dkp-catalog-applications which just points to native github urls in our docs (and support portal too i think).

Of course we cand upload dkp-catalog-applications bundle to downloads.mesosphere.io bucket but i think keeping them closer to github would make UX more intuitive (vs downloading some random bundle from s3).

wdyt @gracedo @makkes ?

gracedo: @cgawron1 any thoughts here?

cgawron1: Since this is a bundle. It would be better served in downloads.d2iq.com (aka downloads.mesosphere.io) so as to provide a consistent location across this type of object. Should we consider keeping it in Github, I think it would need to be an asset under releases so that it doesn't accidentally get moved.  Are there any security considerations?

msdolbey:
Other than the above conversation - that looks legit, though.

pages/dkp/kommander/2.1/install/air-gapped/index.md
takirala: > Since this is a bundle.

The word `bundle` is heavily used here probably. It is just the source code of a (public) repository in archive format.

> Should we consider keeping it in Github, I think it would need to be an asset under releases so that it doesn't accidentally get moved.

It is currently an asset. Infact, its not even a manually curated asset, it is literally the source code served in archive format. here - https://github.com/mesosphere/kommander/releases/tag/v2.1.1

> Are there any security considerations?

The repository is public and the file in question is just an archive of source code. I dont think there are any security concerns.

The ONLY advantage of having `github.com` links instead of `downloads.d2iq.com` links would be the user would be more educated about how to change the links if they want (switch to a different branch / tag or see the contents of the archive in a browsable manner, see release notes, bug fixes etc..).

With all the above said, I do agree that it could be inconsistent using `github.com` and `downloads.d2iq.com` in various places but i think it would be worth the "inconsistency".

takirala: We already publish at `downloads.mesosphere.com` so i can change the link if yall think that would be neat(er), there is no additional work.

cmastr:
Will need to backport to develop 2.1 and 2.2.

msdolbey:
I think we did this already in that 2.2 link changey one

pages/dkp/kommander/2.1/install/air-gapped/index.md
takirala: updated to use downloads.mesophere.com bucket

cmastr (approved):

gracedo (approved):

PR #4026, opened 2022-01-11
fix: missing PreprovisionedInventory label for md

dkoshkin:
Adding a missing label reported in https://mesosphere.slack.com/archives/C02MR1P7286/p1641919879062300.

Based it on `main` because this is a bug in existing docs, but let me know if I should PR it somewhere else.


msdolbey (approved):
Good catch @dkoshkin !

👍 

eep

cneth (approved):

PR #4038, opened 2022-01-13
docs: cleaning up yaml for apps area of install

msdolbey:
## Jira Ticket

[D2IQ-83639](https://jira.d2iq.com/browse/D2IQ-83639)

Just cleaning up this YAML to add context of where this is in the install yaml file, and also alphabetizing things on the same level of the YAML file.

## Checklist

### DIFF
--- a/pages/dkp/kommander/2.1/install/air-gapped/index.md
+++ b/pages/dkp/kommander/2.1/install/air-gapped/index.md
@@ -168,10 +168,13 @@ Based on the network latency between the environment of script execution and the
     kommander install --init > install.yaml
     ```
 
-1. Adapt the configuration file for the air-gapped deployment by changing the `.apps.kommander` field. Ensure you  use the actual version number everywhere `${VERSION}` appears.:
+1. Adapt the configuration file for the air-gapped deployment by changing the `.apps.kommander` section. Ensure you use the actual version number everywhere `${VERSION}` appears:
 
     ```yaml
+    apiVersion: config.kommander.mesosphere.io/v1alpha1
+    kind: Installation
     apps:
+      ...
       kommander:
         values: |
           authorizedlister:
@@ -182,9 +185,6 @@ Based on the network latency between the environment of script execution and the
               manager:
                 image:
                   tag: ${VERSION}-amd64
-          webhook:
-            image:
-              tag: ${VERSION}-amd64
           fluxOperator:
             containers:
               manager:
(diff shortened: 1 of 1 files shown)


mwrightd2 (approved):

pages/dkp/kommander/2.1/install/air-gapped/index.md
```diff
@@ -168,10 +168,13 @@ Based on the network latency between the environment of script execution and the
...
     ```
 
-1. Adapt the configuration file for the air-gapped deployment by changing the `.apps.kommander` field. Ensure you  use the actual version number everywhere `${VERSION}` appears.:
+1. Adapt the configuration file for the air-gapped deployment by changing the `.apps.kommander` section. Ensure you use the actual version number everywhere `${VERSION}` appears:
 
     ```yaml
+    apiVersion: config.kommander.mesosphere.io/v1alpha1
+    kind: Installation
     apps:
+      ...
```
takirala, line 177: this just makes it harder to copy paste right? i agree its probably less confusing as the user would see the entire context but if you have ever tried to copy paste partial yaml, this is only making it more fragile.

msdolbey: That's a good point - I guess this is just a response to the request that it isn't as clear.

I don't know the best balance to it, though. 

@cmastr - what are your thoughts?

I will say that the kommander section in the `apps` was not the first one there, so if they wanted to copy/paste anyway, they would have to do some text editor work.

takirala (approved):

pages/dkp/kommander/2.1/install/air-gapped/index.md
takirala: > I will say that the kommander section in the apps was not the first one there, so if they wanted to copy/paste anyway, they would have to do some text editor work.

True. But before this PR, you would only be copy-pasting and hopefully not fixing indent. After this PR is merged, they _most likely_ need to copy paste AND edit indent if they did not copy the right amount "partial content of the code snippet" or paste the snippet at the "right space in a line".

lgtm either way, just wanted to bring this to your attention.

msdolbey: For sure. I did check the indenting - and that should be good, because I didn't indent anything extra (the good thing is `apps` is at the same level, so that kind of made everything easier). But good point. I definitely don't think about that every time.

msdolbey: Did talk to Tarun during standup - and this is looking decent. Will likely refactor in 2.2.

PR #4072, opened 2022-01-24
add D2IQ-83994 requested redirects

fatz:
## Jira Ticket
https://jira.d2iq.com/browse/D2IQ-83994

added requested redirects

## Checklist


fatz:
ok s3 bucket redirects are limited... I need to find a better solution

cneth (approved):
Overall, LGTM.

That said, perhaps I'm missing something, but is there any reason why we can't just have one redirect for each major DC/OS version instead of redirecting all the internal pages?

I.e. Wouldn't this:

{ "Condition": { "KeyPrefixEquals": "/mesosphere/dcos/1.9/" }, "Redirect": { "HostName": "$REDIR_HOSTNAME", "ReplaceKeyPrefixWith": "mesosphere/dcos/", "HttpRedirectCode": "307" } },

Cover *all* the 1.9 links?   It definitely would make reading this file less confusing and guarantee we hadn't missed any pages.

mwrightd2 (approved):

cmastr (approved):

vacas5 (changes requested):
@cmastr I am pretty sure we are limited to 50 redirects, so I don't think this strategy will work.

msdolbey (changes requested):
Yes - I believe this won't even build/deploy because we're limited at 50 re-directs using this method. 

We could add some other redirect method, but it couldn't be this way.

msdolbey:
As long as we don't care the way that we do this - I know there's a simple metalsmith re-dir that we could take advantage of. If we don't care about prefix stuff, and just ... the redirect to a static page, then this is doable that way? I can make a PR off of this one tomorrow or something

msdolbey:
Oh, nm, I see you already have that in the ticket, Jan, don't mean to step on toes or anything here.

cmastr:
This is just a test from Jan. This request came from Marketing. I understand now that we have a 50 limit and we need to figure that out. To @cneth point, we may just want a strategy for ALL of a given version of a product's docs that is deprecated to go back to the main docs site. If there's a way to implement 'one redirect to rule them all' I'm open to that.

mwrightd2:
@cmastr is there anything I can help with on this ticket to keep it pushing?

PR #4111, opened 2022-02-09
updated 1.21 to 1.22

gvasquezvargas:
Jira: https://jira.d2iq.com/browse/D2IQ-82431?filter=57924
Updated to 1.22 in the scope of support.


JohnB64-JBV (approved):

cmastr (approved):

mikolajb (approved):

msdolbey (commented):
Good catch - but we'll have to change this Kommander version to 2.2, because kommander 2.1 doesn't support k8s to 1.22, I believe.

PR #4150, opened 2022-03-01
Gvv cops 7039 refer to deployed cluster

gvasquezvargas:
Jira: https://jira.d2iq.com/browse/COPS-7039?filter=57924

Customers are overlooking that `kubectl` configuration must point to the correct cluster before running the `install` command. Goal here is to provide with alternatives on how to install Kommander on the correct cluster (and not accidentally installing it on bootstrap cluster). 
PS: Ignore the third file, I was testing stuff and did not want to include that here, my bad!

### Preview

http://docs-d2iq-com-pr-4150.s3-website-us-west-2.amazonaws.com/


pages/dkp/kommander/2.1/release-notes/index.md
```diff
@@ -1,11 +1,11 @@
 ---
 layout: layout.pug
```
cmastr, line 2: Something's off here with this file. We should make sure we remove this from the PR before we merge it.

cmastr:
Let's add this to the 2.2 documentation as well please.

pages/dkp/kommander/2.0/install/networked/index.md
```diff
@@ -66,9 +66,11 @@ If you are installing Kommander on kind, you must know the following:
     ```
 -->
 
-## Install on Konvoy
+## Install Kommander on Konvoy
 
-Before running the commands below make sure that your `kubectl` configuration is pointing to the cluster you want to install Kommander on by setting the `KUBECONFIG` environment variable to the respective kubeconfig file's location.
+Before running the commands below, ensure that your `kubectl` configuration **references the cluster on which you want to install Kommander**, otherwise it will install on the bootstrap cluster. You can do this by setting the `KUBECONFIG` environment variable [to the appropriate kubeconfig file's location][k8s-access-to-clusters].
```
Colin1841, line 71:
```suggestion
Before running these commands, ensure that your `kubectl` configuration **references the cluster where you want to install Kommander**, otherwise it installs on the bootstrap cluster. Do this by setting the `KUBECONFIG` environment variable [to the appropriate kubeconfig file's location][k8s-access-to-clusters].
```

Colin1841: Try to avoid using above or below like this, because sometimes we move sections around and then the order of things is off. It's understood that the verbiage here is the lead-in for the commands, so no need to specify which commands.
"will" with a verb is almost always indicator of passive language, so make it active by removing the "will".
Removed "you can", since it is understood - it's not wrong, but take any opportunity you can to reduce how many words needed to say the same thing.

pages/dkp/kommander/2.1/install/networked/index.md
```diff
@@ -41,9 +41,11 @@ More information on setting a StorageClass as default can be found at [Changing
 
 ## Install Kommander on Konvoy
 
-Before running the commands below, ensure that your `kubectl` configuration references the cluster on which you want to install Kommander. You can do this by setting the `KUBECONFIG` environment variable to the appropriate kubeconfig file's location.
+If you want to customize a Kommander installation, see the [configuration page][configuration-kommander] for more details.
```
Colin1841, line 44:
```suggestion
To customize a Kommander installation, see the [configuration page][configuration-kommander] for more details.
```

pages/dkp/kommander/2.1/install/networked/index.md
```diff
@@ -41,9 +41,11 @@ More information on setting a StorageClass as default can be found at [Changing
...
 
-<p class="message--note"><strong>NOTE:</strong> See the <a href="../configuration">configuration</a> page for more details on how to customize a Kommander installation.</p>
+Before running the commands below, ensure that your `kubectl` configuration **references the cluster on which you want to install Kommander**, otherwise it will install on the bootstrap cluster. You can do this by setting the `KUBECONFIG` environment variable [to the appropriate kubeconfig file's location][k8s-access-to-clusters].
```
Colin1841, line 46: copy edits down from above.

Colin1841 (approved):

cneth (changes requested):
The changes to the install pages look fine, but this PR should NOT be merged as it, because it deletes the 2.1 release notes page.

I guess perhaps an inadvertent change was added to this file, but deleting the file is the wrong fix.    There are many different ways to fix this, but my suggestion would be to just close this PR, create a new branch, and reapply the changes to it and then submit a new PR.

PR #4152, opened 2022-03-01
Qa/remove copy from shell code snippets

msdolbey:

The gist is we can remove the copy icon from code snippets that are just responses/and shouldn't be copied.
SO, this will make it so that any code snippet fences that have `sh` will not show that copy icon.
BUT, so we still make sure that the copy is there for other types of code snippets, just reviewing all the pages that have code snippets and editing those that have `sh` for executable code snippets - and changing it so they are `bash` for the code snippet.

### Preview

http://docs-d2iq-com-pr-4152.s3-website-us-west-2.amazonaws.com/

## Checklist


scss/components/_prism.scss
```diff
@@ -77,3 +77,8 @@ code.language-markdown {
   padding-left: 18px;
   color: #fff;
 }
+
+.code-toolbar pre[class*="language-sh"] {
+  z-index: 1000000;
```
vacas5, line 82: `z-index: 10` has the same outcome. Also, a code comment as to why this is here would be really helpful.

vacas5 (changes requested):

vacas5 (dismissed):
@msdolbey I think we need to add something to the Contributing doc about this if we really want it to be a pattern

cmastr:
@msdolbey can you un-merge conflict this so we can merge it?

cmastr (approved):

mwrightd2 (approved):

PR #4154, opened 2022-03-02
[Konvoy 1.7, 1.8] COPS-7156 HA Control Plane

Colin1841:
Jira: https://jira.d2iq.com/browse/COPS-7156

Added note to 1.7 and 1.8. Please verify if we need to add this to 2.x anywhere

### DIFF
--- a/pages/dkp/konvoy/1.7/networking/control-plane-high-availability/index.md
+++ b/pages/dkp/konvoy/1.7/networking/control-plane-high-availability/index.md
@@ -64,7 +64,7 @@ spec:
         vrid: 51           # optional
 ```
 
-<p class="message--note"><strong>NOTE: </strong>If you run multiple clusters on the same L2 network, you must specify a unique VRID value for each cluster to prevent network issues. In the above example, only use 51 one time, and then choose another value for each additional clusters.</p>
+<p class="message--note"><strong>NOTE: </strong>If you run multiple clusters on the same L2 network, you must specify a unique VRID value for each cluster to prevent network issues. In the above example, only use 51 one time, and then choose another value for each additional cluster.</p>
 
 The IP address specified in `spec.kubernetes.controlPlane.controlPlaneEndpointOverride` is used for the Keepalived VIP. This value is optional if it is already specified in `inventory.yaml` as part of `all.vars.control_plane_endpoint`. You can set `spec.kubernetes.controlPlane.keepalived.interface` to specify the network interface for the Keepalived VIP. This field is optional. If not set, Konvoy automatically detects the network interface to use based on the route to the VIP.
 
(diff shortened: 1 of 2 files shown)


cmastr (approved):

pages/dkp/konvoy/1.8/networking/control-plane-high-availability/index.md
```diff
@@ -64,13 +64,11 @@ spec:
         vrid: 51           # optional
 ```
 
-The IP address specified in `spec.kubernetes.controlPlane.controlPlaneEndpointOverride` is used for the Keepalived VIP.
-This value is optional if it is already specified in `inventory.yaml` as part of `all.vars.control_plane_endpoint`.
-You can set `spec.kubernetes.controlPlane.keepalived.interface` to specify the network interface for the Keepalived VIP.
-This field is optional. If not set, Konvoy automatically detects the network interface to use based on the route to the VIP.
+<p class="message--note"><strong>NOTE: </strong>If you run multiple clusters on the same L2 network, you must specify a unique VRID value for each cluster to prevent network issues. In the above example, only use 51 one time, and then choose another value for each additional clusters.</p>
```
Colin1841, line 67:
```suggestion
<p class="message--note"><strong>NOTE: </strong>If you run multiple clusters on the same L2 network, you must specify a unique VRID value for each cluster to prevent network issues. In the above example, only use 51 one time, and then choose another value for each additional cluster.</p>
```

pages/dkp/konvoy/1.7/networking/control-plane-high-availability/index.md
```diff
@@ -64,13 +64,11 @@ spec:
         vrid: 51           # optional
 ```
 
-The IP address specified in `spec.kubernetes.controlPlane.controlPlaneEndpointOverride` is used for the Keepalived VIP.
-This value is optional if it is already specified in `inventory.yaml` as part of `all.vars.control_plane_endpoint`.
-You can set `spec.kubernetes.controlPlane.keepalived.interface` to specify the network interface for the Keepalived VIP.
-This field is optional. If not set, Konvoy automatically detects the network interface to use based on the route to the VIP.
+<p class="message--note"><strong>NOTE: </strong>If you run multiple clusters on the same L2 network, you must specify a unique VRID value for each cluster to prevent network issues. In the above example, only use 51 one time, and then choose another value for each additional clusters.</p>
```
Colin1841, line 67:
```suggestion
<p class="message--note"><strong>NOTE: </strong>If you run multiple clusters on the same L2 network, you must specify a unique VRID value for each cluster to prevent network issues. In the above example, only use 51 one time, and then choose another value for each additional cluster.</p>
```

cneth (approved):

PR #4161, opened 2022-03-08
Add Control Plane High Availability to Kommander 2.1 and 2.2

lukeogg:
## Jira Ticket

https://jira.d2iq.com/browse/D2IQ-78944

Add the networking/Control Plane High Availability from Konvoy 1.8 to Kommander 2.1 and 2.2. This document appears to still be relevant to the newer versions.

I am adding it to the Kommander documentation as there is a Networking topic there that already references Konvoy. This is with the view that the documentation will be rearranged to 

### Preview

http://docs-d2iq-com-pr-4161.s3-website-us-west-2.amazonaws.com/

### DIFF
--- a/pages/dkp/kommander/2.1/networking/control-plane-high-availability/index.md
+++ b/pages/dkp/kommander/2.1/networking/control-plane-high-availability/index.md
@@ -1,0 +1,111 @@
+---
+layout: layout.pug
+navigationTitle: Control Plane High Availability
+title: Control Plane High Availability
+menuWeight: 8
+excerpt: Make Konvoy control plane nodes highly available
+beta: false
+enterprise: false
+---
+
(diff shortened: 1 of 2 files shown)


cneth (changes requested):
I am confused why we think we need this page in the Kommander docs...   The Control plane is a Konvoy topic only, there is nothing in Kommander that deals with the control plane configuration.   

We don't have the clusterprovisioner in DKP 2.x, so that needs a rewrite.
Also, we don't use keepalived in DKP2.x, so that section would need to be rewritten.

We have Control plane info in the install/preprovisioned area in the Konvoy 2.x docs, which covers most of this.

If we feel we need a topic like this, it should be in the Konvoy docs only, and probably should be part of the 'choose infrastructure' section although much of it would be the same for all the providers.

PR #4189, opened 2022-03-18
COPS-7172 added URL

Colin1841:
https://jira.d2iq.com/browse/COPS-7172

Added a page-to-page URL link in both 2.1 and 2.2.

### DIFF
--- a/pages/dkp/konvoy/2.1/choose-infrastructure/pre-provisioned/define-control-plane-endpoint/index.md
+++ b/pages/dkp/konvoy/2.1/choose-infrastructure/pre-provisioned/define-control-plane-endpoint/index.md
@@ -31,7 +31,7 @@ If an external load balancer is not available, use the [built-in virtual IP](../
 
 ## Single-Node control plane
 
-A control plane with one node can use its single node as the endpoint, so you will not require an external load balancer, or a built-in virtual IP. However, do not use a single-node control plane configuration in production, as it is not possible to upgrade the control plane if you only have one node. This is due to the fact that CAPI does not perform in-situ upgrades, instead, it replaces machines. If you want to upgrade your cluster, use the default `<code>`--control-plane-replicas`</code>` setting of 3.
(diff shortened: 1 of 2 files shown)


pages/dkp/konvoy/2.1/choose-infrastructure/pre-provisioned/define-control-plane-endpoint/index.md
```diff
@@ -28,11 +27,11 @@ It is recommended that an external load balancer be the control plane endpoint.
...
 
 ## Single-Node control plane
 
-A control plane with one node can use its single node as the endpoint, so you will not require an external load balancer, or a built-in virtual IP. However, do not use a single-node control plane configuration in production, as it is not possible to upgrade the control plane if you only have one node. This is due to the fact that CAPI does not perform in-situ upgrades, instead, it replaces machines. If you want to upgrade your cluster, use the default <code>--control-plane-replicas</code> setting of 3.
+A control plane with one node can use its single node as the endpoint, so you will not require an external load balancer, or a built-in virtual IP. However, do not use a single-node control plane configuration in production, as it is not possible to upgrade the control plane if you only have one node. This is due to the fact that CAPI does not perform in-situ upgrades, instead, it replaces machines. If you want to upgrade your cluster, use the default `<code>`--control-plane-replicas`</code>` setting of 3.
```
cmastr, line 34:
```suggestion
A control plane with one node can use its single node as the endpoint, so you will not require an external load balancer, or a built-in virtual IP. However, do not use a single-node control plane configuration in production, as it is impossible to upgrade the control plane if you only have one node. This is because CAPI does not perform in-situ upgrades, instead, it replaces machines. If you want to upgrade your cluster, use the default `<code>`--control-plane-replicas`</code>` setting of 3.
```

cmastr (approved):

cneth (approved):

PR #4216, opened 2022-03-28
fix: updates to support arm64 machines

nataliepina:
## Jira Ticket

I have been troubleshooting some of the issues I've seen while working on an M1 arm64 machine. 

Issues and reasons for fixes:
1. node-sass. There is a warning when you try to run  `npm run dev` on an arm64 machine that will prevent you from running. The fix for this was found through this thread - https://stackoverflow.com/questions/68095626/node-sass-with-apple-m1-big-sur-and-arm64. This led to the required change to add calc() around the division math. 
2. Suggested node version warnings. I've bumped the suggested node version. This means we'll need to make sure everyone is on version node 16 or higher.
3. Dockerfile will need to use `FROM node:gallium-buster-slim` which has facilitates `make docker-liveedit-image` on arm64

It would probably be helpful to test out all of these commands on a non-arm64 to confirm the changes. 

The only warning I am still seeing is:
```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'shortcode-parser@0.0.1',
npm WARN EBADENGINE   required: { node: '0.10.x' },
npm WARN EBADENGINE   current: { node: 'v16.14.0', npm: '8.3.1' }
npm WARN EBADENGINE }
```
However, shortcode-parser is already on the latest version possible. 
After further investigation. The shortcode-parser package.json actually has that set as the required version. See here: https://github.com/mendezcode/shortcode-parser/blob/master/package.json

I submitted an issue to them for this. It looks like this is a pretty old repo though (last published 8 years ago) so we might be best to wait for an alternative library to integrate with future upgrades. 

### Preview

http://docs-d2iq-com-pr-<add_pr_##_here>.s3-website-us-west-2.amazonaws.com/

## Checklist

- [x] Test all commands and procedures, if applicable.


docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:lts-buster-slim
+FROM node:fermium-buster-slim
```
mstruebing, line 1: Not using hardcoded LTS is fine but fermium is the codename of node v14 whereas in `package.json` `engines` you say `^16.0.0` which would be the codename Gallium, see: https://nodejs.org/en/about/releases/

So we should use it here too - if that works.

To your question: we should definitely commit this as this defines which node version we use in the docker image and this should be the same as locally.
There is another Docker image which we should align: `docker/Dockerfile.production`
This should be using the same base image I think.

package.json
```diff
@@ -62,12 +61,13 @@
     "eslint-config-airbnb": "^19.0.2",
     "html-entities": "^2.1.0",
     "mustache": "^4.1.0",
+    "node-sass": "npm:sass@^1.49.9",
```
mstruebing, line 64: What a good find 🥇

package.json
```diff
@@ -62,12 +61,13 @@
...
     "react": "^17.0.1",
     "react-dom": "^17.0.1"
   },
   "engines": {
-    "node": "^12.0.0",
-    "npm": "^6.0.0"
+    "node": "^16.0.0",
+    "npm": ">=7.0.0"
```
mstruebing, line 70: Just as a note, I think this means people working on the docs site need to update their local environments.

mstruebing: When using this versions (editing the `.node-version` file) I get this warning: 

```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'shortcode-parser@0.0.1',
npm WARN EBADENGINE   required: { node: '0.10.x' },
npm WARN EBADENGINE   current: { node: 'v16.14.2', npm: '8.5.0' }
npm WARN EBADENGINE }
```

I'm not sure this is a problem, I was able to run the site without problems.

mstruebing (commented):
There is a `.node-version` which we need to adjust to the new version too because multiple tools are looking for a `.node-version` file and determine which version to use based on this file, which is still using v12.

I was able to run the site locally and in the live edit image too, but we really should align these versions as this will lead to further problems in the future I think.

package.json
nataliepina: I was seeing the same thing with shortcode-parser actually but in the last few npm installs I hadn't seen it again. I tried to update the version for it, but it's already on the latest possible.

docker/Dockerfile
nataliepina: Thanks for this info! I wasn't 100% sure what we should set this to for it to work for everyone. I will test out some updated options.

nataliepina: Good call though, gallium is working for me. I've updated both Dockerfile locations.

README.md
```diff
@@ -37,8 +37,8 @@ If you do not want to setup `node.js` on your machine and instead have docker in
 MOUNT_SRC=$PWD/docs/site/
 
 # where in the url-structure to attach the folder you just selected. without the leading slash.
-# e.g. for konvoy this might be "dkp/konvoy/1.7".
-MOUNT_DST=dkp/konvoy/1.7
+# e.g. for kommander this might be "dkp/konvoy/2.2".
+MOUNT_DST=dkp/kommander/2.2
```
vacas5, line 41: I think we can leave this as it was. It will always be a moving target

vacas5 (changes requested):
there is a node-version file that should be updated to match the new desired node version

README.md
nataliepina: The previous command did not work for me at all though, I received an error, so in case someone tries to do a copypasta I wanted to update. I don't know if there is an issue with the use of 1.7 directly.

mstruebing (approved):
This looks good from my point of view - good work 👍 

But we should wait with merging: https://mesosphere.slack.com/archives/C012B0X8JLA/p1648587060937769

seialkali (approved):
I've tested this on my M1 mac and it's working beautifully. Great work fixing this @nataliepina 🎉

gvasquezvargas (approved):
Since this seems to be code only, and probably you need docs approval to merge, approved!

nataliepina:
@vacas5  Are we safe to merge this one yet? We can be sure everyone is on the updated version of node/npm and help to support that transition.

Colin1841 (approved):

vacas5 (changes requested):
Sorry @nataliepina , but the readme defines which node version to use. Can you update that as well as a part of this?

nataliepina:
> Sorry @nataliepina , but the readme defines which node version to use. Can you update that as well as a part of this?

No problem! Good catch. Updated ✔️

vacas5 (approved):

PR #4309, opened 2022-04-08
Add Kaptain 2.0.0 Docs (Moved from Kaptain repo)

juliangieseke:
ℹ️ This PR/branch is considered reviewed at this point, its too big to review but everything was in individual PRs (see below) and also its disabled by config, so merging this should not have any impact on the docs page.

🔍 Preview Link: http://docs-d2iq-com-pr-4309.s3-website-us-west-2.amazonaws.com/dkp/kaptain/2.0.0/

### General Info
* Kaptain Docs are hidden via config, but we still can wait to merge this until DKP 2.2 is GA. (AFAIK #4266 ?)
* There are individual PRs per folder in order to make them reviewable and also not to overwhelm vale.. seems like some folders have issues.
* This PR should be squashed and its commit message should have a ref to https://github.com/mesosphere/kaptain/tree/b0fb5a203e115de87ef9e341d1ed81b409f75281/docs/site (where these files are copied from).
* Main goal is to get this merged ASAP to unblock further work, it doesnt has to be perfect or even good, we can improve later (its hidden after all). so I'd only focus on blocking CI issues (maybe vale errors?)

JIRA: [D2IQ-87565](https://jira.d2iq.com/browse/D2IQ-87565)

PR #4325, opened 2022-04-10
Kaptain 2.0.0 Docs user-management folder

juliangieseke:
Kaptain 2.0.0 docs "child PR" for #4309 (check its description)

### DIFF
--- a/pages/dkp/kaptain/2.0.0/user-management/index.md
+++ b/pages/dkp/kaptain/2.0.0/user-management/index.md
@@ -1,0 +1,249 @@
+---
+layout: layout.pug
+navigationTitle: Manage Users
+title: Manage Users and Permissions
+menuWeight: 20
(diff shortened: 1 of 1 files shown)


juliangieseke:
vale complains about wording/grammar?!

gvasquezvargas:
Vale complains about:
-Kommander not being capitalized in commands/paths (which does not make sense because we don't capitalize in those cases)
-Onboarding being capitalized (which does not make sense here since it is the first word of a title/header and SHOULD be capitalized.
Looks to me as Vale is complaining more than usual, @cmastr I remember you telling me that something was changed in the docs repo, so Vale is a bit less picky. Maybe we need to apply those changes to Kaptain folders being added manually?

juliangieseke:
yeah, such "global" switch would be nice given its complaining about such things in multiple PRs.

PR #4376, opened 2022-04-21
rename kaptain folders (e.g. 1.3.0 -> 1.3)

juliangieseke:
## Jira Ticket

No JIRA

Noticed that Kaptain is the only Product using x.y.z folder names, this aligns it with other products. 
Updated config, added redirects and checked for dead links (found 0). So this should cover everything? 

### Preview

http://docs-d2iq-com-pr-4376.s3-website-us-west-2.amazonaws.com/

## Checklist


juliangieseke:
hm, this PR has no changes, only renamed.. vale complaining is unrelated.

Colin1841:
@juliangieseke - this could break internal hyperlinks, from page to page within the docs set. For example, this changes the page title, so, therefore, changes the URL for this page: 
[pages/dkp/kaptain**/1.3.0/**index.md → pages/dkp/kaptain**/1.3/**index.md]
We recommend leaving these as-is in current versions, but we can make the change going forward in 2.0 if you want/need to.

juliangieseke:
@Colin1841 ah ok, I thought by adding these redirects such things would be covered: https://github.com/mesosphere/dcos-docs-site/pull/4376/files#diff-5b55699c9a670dffc128a49c6973831ee15e6759de6f20f6fcd03cbf5a475cd3

but yeah, if there is a risk of breaking something I agree we shouldnt do this, also given the confluence move this might not be as important in the first place, so we could simply stick to x.y.z for the time being.. WDYT?

PR #4401, opened 2022-04-26
Updated Minimum Requirements for Kaptain Installation. 

andresmascl:
Please do see https://docs.google.com/spreadsheets/d/1VoF4MCsiUD2NnogcPZ-2wSsco1auZY-j91o_nHZPxMY/edit?usp=sharing for supporting documentation

Helps fix https://jira.d2iq.com/browse/D2IQ-83732.  Minimum requirements for GPU assisted deployments is still WIP

http://docs-d2iq-com-pr-4401.s3-website-us-west-2.amazonaws.com/

### DIFF
--- a/pages/dkp/kaptain/2.0.0/install/dkp/index.md
+++ b/pages/dkp/kaptain/2.0.0/install/dkp/index.md
@@ -14,18 +14,11 @@ You can deploy Kaptain to a cluster in a selected workspace. If you do not inten
 
 ## Requirements
 
-Before proceeding, verify that your environment meets the following basic requirements:
+Before proceeding, verify that your cluster has freely available the following resources:
 
-- Control plane
-  - min. 3 nodes
-  - min. 4 cores _per node_
-  - min. 200 GiB free disk space _per node_
-  - min. 16 GiB RAM _per node_
-- Workers
-  - min. 6 nodes
-  - min. 8 cores _per node_
-  - min. 200 GiB free disk space _per node_
-  - min. 32 GiB RAM _per node_
+- 4 CPU cores
+- 16 GiB of RAM
+- 170 GiB or Storage Memory
 - GPUs (optional)
   - NVIDIA only
   - min. 200 GiB free disk space per instance


chessahoekstra (commented):
why are there no more node counts?

andresmascl:
> 

As I understand nodes are a requirement for Kommander and not for Kaptain

pages/dkp/kaptain/2.0.0/install/dkp/index.md
```diff
@@ -14,18 +14,11 @@ You can deploy Kaptain to a cluster in a selected workspace. If you do not inten
...
-- Workers
-  - min. 6 nodes
-  - min. 8 cores _per node_
-  - min. 200 GiB free disk space _per node_
-  - min. 32 GiB RAM _per node_
+Before proceeding, verify that your cluster has freely available the following resources:
+
+- 4 CPU cores
+- 16 GiB of RAM
+- 170 GiB or Storage Memory
```
juliangieseke, line 21: is this 
```suggestion
- 170 GiB of Storage
```
or 
```suggestion
- 170 GiB of Memory
```
?

https://v0-6.kubeflow.org/docs/started/k8s/overview/

gvasquezvargas:
Hi all! @mstruebing and I realized that the checks were failing in this PR and think this might be due to the fact that @andresmascl is a contractor and therefore using an external fork. In this case, we think it would be best to have one of the admins force merging once this PR is ready without the checks passing. 
But it might be a good idea for us to request access for him? WDYT @juliangieseke ?

alembiewski (changes requested):
Thanks for updating the requirements, @andresmascl! Few comments:
- As @chessahoekstra mentioned, the worker node count is still a valid requirement and it should be no less than 3 by default (for the fault tolerance and even distribution of the MySQL cluster data nodes).
- I trying to understand CPU measurements here: https://docs.google.com/spreadsheets/d/1VoF4MCsiUD2NnogcPZ-2wSsco1auZY-j91o_nHZPxMY/edit?usp=sharing. Why is it showing GiB for CPU graphs? I guess what it shows is the average number of CPU cores used during the last 5 minutes, and it would be good to see separate graphs for each namespace (kommander, knative, istio, kubeflow).

In general, I think we should at least add a warning here that these requirements are only for installation and needs to be adjusted depending on the workload.
cc @gvasquezvargas

andresmascl:
> Thanks for updating the requirements, @andresmascl! Few comments:
> 
> * As @chessahoekstra mentioned, the worker node count is still a valid requirement and it should be no less than 3 by default (for the fault tolerance and even distribution of the MySQL cluster data nodes).
> * I trying to understand CPU measurements here: https://docs.google.com/spreadsheets/d/1VoF4MCsiUD2NnogcPZ-2wSsco1auZY-j91o_nHZPxMY/edit?usp=sharing. Why is it showing GiB for CPU graphs? I guess what it shows is the average number of CPU cores used during the last 5 minutes, and it would be good to see separate graphs for each namespace (kommander, knative, istio, kubeflow).
> 
> In general, I think we should at least add a warning here that these requirements are only for installation and needs to be adjusted depending on the workload. cc @gvasquezvargas

Thank you @alembiewski.  I was following [kubeflows' own minimum requirements](https://v0-6.kubeflow.org/docs/started/k8s/overview/).  That is where I got the idea of leaving our the nodes count.  Adding them now.

Regarding CPU measurements:  yes.  the "GiB" slipped through from the RAM measurement image.  Please see the "units" column, where they are specified correctly as "seconds" and "number of cores".

gvasquezvargas:
> In general, I think we should at least add a warning here that these requirements are only for installation and needs to be adjusted depending on the workload. cc @gvasquezvargas
@alembiewski we have that information already right after what Andrés edited. Do you think we should turn that information into a warning?

pages/dkp/kaptain/2.0.0/install/dkp/index.md
```diff
@@ -14,23 +14,13 @@ You can deploy Kaptain to a cluster in a selected workspace. If you do not inten
...
-  - min. 6 nodes
-  - min. 8 cores _per node_
-  - min. 200 GiB free disk space _per node_
-  - min. 32 GiB RAM _per node_
-- GPUs (optional)
-  - NVIDIA only
-  - min. 200 GiB free disk space per instance
-  - min. 64 GiB RAM per instance
-  - min. 12 GiB GPU RAM per instance
+Before proceeding, verify that your cluster has freely available the following resources:
```
Colin1841, line 17:
```suggestion
Before proceeding, verify that your cluster has the following resources freely available:
```

Colin1841 (approved):

alembiewski (approved):
LGTM

timcharper (approved):

pages/dkp/kaptain/2.0.0/install/dkp/index.md
```diff
@@ -14,23 +14,13 @@ You can deploy Kaptain to a cluster in a selected workspace. If you do not inten
...
-  - min. 32 GiB RAM _per node_
-- GPUs (optional)
-  - NVIDIA only
-  - min. 200 GiB free disk space per instance
-  - min. 64 GiB RAM per instance
-  - min. 12 GiB GPU RAM per instance
+Before proceeding, verify that your cluster has the following resources freely available:
+
+- 3 worker nodes
+- 4 CPU cores
```
chessahoekstra, line 20: so these are not requirements per node?

chessahoekstra: if there are still per-node reqs, let's update them but keep the format of the previous

alembiewski:
@andresmascl, I think we might need to adjust these numbers and use different metrics to calculate the minimum requirements. We need to use container resource requests instead of actual consumption because these are what kube-scheduler looks for when choosing a node to schedule a pod. Here are the examples for memory and CPU :

Screen Shot 2022-04-26 at 10 16 37 PM

Screen Shot 2022-04-26 at 10 16 11 PM

pages/dkp/kaptain/2.0.0/install/dkp/index.md
```diff
@@ -14,23 +14,13 @@ You can deploy Kaptain to a cluster in a selected workspace. If you do not inten
...
+- 170 GiB of Storage
+- 16 GiB of RAM
+- NVIDIA GPU instances are supported only
```
alembiewski, line 23:
```suggestion
Before proceeding, verify that your cluster has the following resources freely available:

- Cluster consists of at least 3 worker nodes
- 25 CPU cores
- 32 GiB of RAM
- 170 GiB of Storage
- NVIDIA GPU instances are supported only
```

cmastr: committing this so we can kick off the PR preview.

PR #4539, opened 2022-06-13
Quick Component Update to release Notes.

Colin1841:
No Jira, just a quick fix.

### Preview

http://docs-d2iq-com-pr-4539.s3-website-us-west-2.amazonaws.com/


cneth (approved):

cmastr (approved):

cmastr:
Do you want this PR to cover both Konvoy and Kommander 2.2?

some-things (approved):
LGTM. However, do we know how/why we missed updating the component versions here? I haven't verified, but I would hazard a guess that there are other components in this list with the wrong versions. Is it possible that we're only updating this list for minor releases and not patch releases?

PR #4653, opened 2022-10-24
fixing steps to address NOTICES.txt

pwlmesos:
### Preview

http://docs-d2iq-com-pr-<add_pr_##_here>.s3-website-us-west-2.amazonaws.com/

## Checklist

### DIFF
--- a/pages/dkp/kommander/2.2/install/air-gapped/index.md
+++ b/pages/dkp/kommander/2.2/install/air-gapped/index.md
@@ -84,7 +84,14 @@ Check the built-in help text for each command for more information.
 
 ### Load the Docker images into your Docker registry
 
-1.  See the `NOTICES.txt` file for 3rd party software attributions and place the `kommander-image-bundle-v2.2.2.tar.gz` and `dkp-catalog-applications-image-bundle-v2.2.2.tar.gz` bundles within a location where you can load and push the images to your private Docker registry.
+1. Untar the kommander image bundle
+
+```bash
+tar -xf dkp-catalog-applications-image-bundle-v2.2.1.tar.gz
+```
+
+See the `NOTICES.txt` file for 3rd party software attributions and place the `kommander-image-bundle-v2.2.1.tar.gz` and `dkp-catalog-applications-image-bundle-v2.2.1.tar.gz` bundles within a location where you can load and push the images to your private Docker registry.
+
 
 1.  Run the following command to load the air-gapped image bundle into your private Docker registry:
 


pages/dkp/kommander/2.2/install/air-gapped/index.md
```diff
@@ -84,7 +84,14 @@ Check the built-in help text for each command for more information.
 
 ### Load the Docker images into your Docker registry
 
-1.  See the `NOTICES.txt` file for 3rd party software attributions and place the `kommander-image-bundle-v2.2.2.tar.gz` and `dkp-catalog-applications-image-bundle-v2.2.2.tar.gz` bundles within a location where you can load and push the images to your private Docker registry.
+1. Untar the kommander image bundle
+
+```bash
+tar -xf dkp-catalog-applications-image-bundle-v2.2.1.tar.gz
```
cneth, line 90: @pwlmesos 

We have a separate page already that addresses both the download *and* untarring of the 'dkp-catalog-applications' bundle:   

https://archive-docs.d2iq.com/dkp/kommander/2.2/install/air-gapped/catalog/#load-the-docker-images-into-your-docker-registry

So I am not sure this page (which does not mention catalog applications) should be mentioning them here again.  I think this page should just focus on the 'kommander-image-bundle' file in this section - if you're not installing catalog apps you don't need to push them here.    

Nonetheless, there are definitely issues here - the catalog page has a different problem which is that  someone has removed the .gz suffix from the wget so it will fail a different way, and we don't tell you anywhere on this page where/how to download any of the files mentioned here.   😢 

However, if you *did* follow the instructions on that other page - your command here (without any context) will break the install because you'll have docker/registry/xxx files all over.. so we can't just add this here like this.   

My guess is that folks in your team are running into problems because you don't do the 'wget | tar xvf -' which is the 'documented' way to download these files - presumably you are using the 'miller bundle' which just has the 'raw downloads' (and thus need to be run through the extraction again).   It doesn't help that the inside and the outside file both have the same name...   I think we should probably add some instructions that have you confirm the contents of the bundle file instead of just telling you to 'tar xvf'.

Anyways, we need to clean all this up at once rather than piecemeal.   I'll put a PR up later today that tries to address both pages.

cneth (changes requested):

pages/dkp/kommander/2.2/install/air-gapped/index.md
pwlmesos: If a user did that page already then most of this page is duplicative anyway.  

well then this page should be severely shortened, or entirely replaced with the other page. 

and at this point I a more concerned with the new docs as we shouldn't have fresh installs of 2.2, just closing out this naggingf JIRA

supershal:
@pwlmesos this PR does not seem related to https://d2iq.atlassian.net/browse/D2IQ-93657 as per PR description. can you please confirm it and correct the reference if thats not a case.

cmastr:
removed reference to JIRA.

